PR #2, opened 2016-07-06
KUZBO-129 Authenticated session

jenow:
- Implemented Login with cookie
- Wrote unit tests

TODO:
Write e2e tests
Implement material design when the global-layout branch is finished


src/services/cookies.js
```diff
@@ -0,0 +1,11 @@
+export default {
```
xbill82: I have nothing against cookies, but @AnthonySendra leveraged the SessionStorage in the Klack App, wouldn't it be better? Or even LocalStorage if we want real persistence...
As far as I understand from [this thread](http://stackoverflow.com/questions/3220660/local-storage-vs-cookies), Cookies are essentially useful when their content must be sent to the server at every request (which is not our case).
That being said, this module is useful and should not be dropped even if we choose another method to implement the session.

test/e2e/specs/test.js
```diff
@@ -4,11 +4,9 @@
 module.exports = {
   'default e2e tests': function (browser) {
     browser
-    .url('http://localhost:8080')
-      .waitForElementVisible('#app', 5000)
-      .assert.elementPresent('.logo')
-      .assert.containsText('h1', 'Hello World!')
-      .assert.elementCount('p', 3)
+    .url('http://localhost:3000')
```
xbill82: Would be nice to get this address from a config file.

test/unit/specs/auth/component.spec.js
```diff
@@ -0,0 +1,10 @@
+import Vue from 'vue'
+import Login from '../../../../src/components/Login'
+
+describe('Login.vue', () => {
+  it('should render correct contents', () => {
```
xbill82: This title is copy pasted from somwhere, I think

test/unit/specs/auth/component.spec.js
```diff
@@ -0,0 +1,10 @@
...
+    expect(Login.data().username).to.be.null
+    expect(Login.data().password).to.be.null
+    expect(Login.vuex.actions.doLogin).to.be.defined
+  })
+})
```
xbill82: So we don't test the rendered UI here? Do we leave it to integration tests?

test/unit/specs/auth/mutations.spec.js
```diff
@@ -0,0 +1,20 @@
+import { expect } from 'chai'
+import { mutations } from '../../../../src/vuex/modules/auth/store'
+
+const { SET_CURRENT_USER, SET_CURRENT_USER_RIGHTS } = mutations
+
+describe('auth mutations test', () => {
+  it('should set the user status', () => {
+    let stat = {user: null}
```
xbill82: Is `stat` instead of `state` on purpose?

xbill82:
Great PR. The tests are just GREAT!

src/components/Login.vue
```diff
@@ -23,7 +23,9 @@ export default {
     }
   },
   vuex: {
-    actions: { doLogin },
+    actions: {
+      doLogin
```
jenow: I agree, I will call `logout` `doLogout`

test/unit/specs/auth/component.spec.js
jenow: We should, I wrote it in the TODO

AnthonySendra:
Ok for me, but the merge must be done on branch develop instead of master ;)

jenow:
New PR to develop branch

PR #7, opened 2016-07-07
Plugins system

xbill82:
This is the place to work on the Kuzzle Backoffice Plugins System.

The Plugins System will enable to install a plugin via the web interface and plug a freshly installed plugin on the fly, without page reload. Plugins will be able to add routes to display dedicated pages, but also to enrich the interface of the standard sections (e.g. add a new button to "track a user").

TODO in this PR
- [x] POC of the dynamic (lazyload) require of a plugin file,
- [x] Execute a plugin `default` function with a context,
- [x] create a Plugins service in charge to load, register, execute and unload the configured plugins,
- [x] load plugins as npm modules,
- [x] POC page plugins


src/services/plugins.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default {
+  loadAll () {
+    pluginList.forEach((pluginName) => {
+      this.loadOne(pluginName)
+    })
+  },
+  loadOne (pluginName) {
+    // require(['../../node_modules/kuzzle-bo-plugin-' + pluginName + '/index'], (plugin) => {
+    require(['../../plugins/kuzzle-bo-plugin-' + pluginName + '/index'], (plugin) => {
```
jenow: Does it really have to being with `kuzzle-bo-plugin` ?

src/vuex/modules/plugins/store.js
```diff
@@ -0,0 +1,20 @@
+import {
+  ADD_ITEM,
+  ADD_GROUP
+} from './mutation-types'
+
+const state = {
+  elements: {}
+}
+
+const mutations = {
```
jenow: We should fill those to show how to really add an item / group of item

src/services/plugins.js
xbill82: It's a way to avoid that Webpack builds the whole `node_modules` (in case the plugins are located there, which is something we should discuss)

src/vuex/modules/plugins/store.js
xbill82: definitely

PR #8, opened 2016-07-07
Global layout

AnthonySendra:
* Add materializecss (sass)
* Add main menu
* Add component Security (just the layout)


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     // allow paren-less arrow functions
     'arrow-parens': 0,
     // allow debugger during development
-    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0
+    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
+    'no-trailing-spaces': 0
```
xbill82: This is the beginning of the end.

docker-compose/run-dev.sh
```diff
@@ -1,25 +1,7 @@
 #!/bin/sh
 
-kuzzle=${KUZZLE_HOST:-kuzzle:7511}
```
xbill82: I don't know if this was useful anymore, but I'm sure it didn't work due to a config bug I discovered this weekend (we make a wrong use of the `webpack.DefinePlugin`). Anyway, if this happens to be useless, let's take it out.

docker-compose/run-test.sh
```diff
@@ -2,20 +2,26 @@
 
 kuzzle=${KUZZLE_HOST:-kuzzle:7511}
 
-while ! curl -m 2 -silent -output /dev/null http://$kuzzle/api/1.0 > /dev/null
```
xbill82: ... but I think we need it to ensure the integration tests to start only once Kuzzle is up.

src/assets/layout.scss
```diff
@@ -0,0 +1,24 @@
+main {
```
xbill82: The Force tells me this should be placed in the `Layout.vue` style tag.

src/main.js
```diff
@@ -2,10 +2,13 @@ import Vue from 'vue'
 import router from './services/router'
 import App from './App'
 import store from './vuex/store'
+import { loginFromCookie } from './vuex/modules/auth/actions'
 
 Vue.config.debug = process.env.NODE_ENV !== 'production'
 
 router.start({
```
xbill82: Shouldn't this be triggered _after_ the loginFromCookie, like, in a callback?

jenow:
+1

PR #17, opened 2016-07-26
Named main components

xbill82:
The main components now have a name, so that we have no more "AnonymousComponent" in the Vue Chrome Extension. [Dad says it in this thread]( http://forum.vuejs.org/topic/792/vue-router-components-show-as-anonymous-component-how-do-i-name-them/2).

In the future, we should try to be consistent with the convention of explicitly setting a `name` property to all our components.


AnthonySendra:
:+1:

jenow:
+1

PR #19, opened 2016-07-27
KUZBO-132 watch data

jenow:


src/components/Materialize/collapsible.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  id: 'collapsible',
+  bind: () => {
+    setTimeout(() => {
+      /* eslint no-undef: 0 */
```
scottinet: This eslint pragma disables the linter for all subsequent lines of the file. From the position of this comment, I believe what you meant was: `// eslint-disable-next-line no-undef`

src/vuex/modules/data/actions.js
```diff
@@ -1,8 +1,10 @@
 import kuzzle from '../../../services/kuzzle'
-import {RECEIVE_MAPPING, RECEIVE_INDEXES_COLLECTIONS} from './mutation-types'
+import {RECEIVE_MAPPING, RECEIVE_INDEXES_COLLECTIONS, ADD_NOTIFICATION, EMPTY_NOTIFICATION} from './mutation-types'
 import {SET_ERROR} from '../common/mutation-types'
 import Promise from 'bluebird'
 
+let room
```
scottinet: As seen together, this global variable should instead be put in the store.

test/unit/specs/data/actions.spec.js
```diff
@@ -8,14 +8,14 @@ describe('listIndexesAndCollections action', () => {
     '../../../services/kuzzle': {
       listIndexes (cb) {
         if (triggerError[0]) {
-          cb({message: 'error'})
+          cb({message: 'error', stack: 'stack'})
```
scottinet: This shouldn't be necessary since bluebird 3.3

src/vuex/modules/data/store.js
```diff
@@ -14,6 +19,15 @@ export const mutations = {
   },
   [RECEIVE_MAPPING] (state, mapping) {
     state.mapping = mapping
+  },
+  [ADD_NOTIFICATION] (state, notification) {
+    state.notifications.push(notification)
```
xbill82: Shouldn't we limit the size of this array somehow?

xbill82:
A part from the conflicts and my comments, :+1:

src/vuex/modules/data/store.js
jenow: I was thinking it could be subject to another PR

stafyniaksacha:
+1

PR #23, opened 2016-07-27
KUZBO-137 Index summary

jenow:
Provides the summary page for a given index.


stafyniaksacha:
+1

jenow:
+1

src/components/Data/Browse/IndexBranch.vue
```diff
@@ -2,7 +2,7 @@
   <div :class="{ 'open': open }">
     <i v-if="collectionCount(index)" class="fa fa-caret-right tree-toggle" aria-hidden="true" @click="toggleBranch()"></i>
     <!-- v-link="{name: 'DataIndex', params: {index: index.name}}" -->
```
xbill82: Let's take this comment out :)

src/vuex/modules/data/store.js
```diff
@@ -1,11 +1,16 @@
...
-  RECEIVE_MAPPING
+  RECEIVE_MAPPING,
+  RECEIVE_COLLECTIONS
 } from './mutation-types'
 
 const state = {
   indexesAndCollections: undefined,
-  mapping: undefined
+  mapping: undefined,
+  collections: {
```
xbill82: **Not to be changed in this PR, but...**
I'm afraid that this naming is too "broad" and that we'll soon be in trouble understanding what `collections` are we talking about... We have the collections that are shown in the Tree View (which are organized by index), then we have this one `collection`, which contains the ones that are shown in the Index Summary page (so they're basically "The List Of Collections Of the Currently Browsed Index", let's find a better name for that).
Then, do we have any other collection list? May we take five minutes to think about it together?

xbill82:
:+1:

PR #24, opened 2016-07-27
Data main vues : indexes create modal

stafyniaksacha:
Add create index modal & interaction with indexes browse


AnthonySendra:
:+1:

src/components/Data/Indexes/ModalCreate.vue
```diff
@@ -0,0 +1,111 @@
+<template>
```
jenow: Why don't we use the Modal component and use props to pass all the variables ?

stafyniaksacha: To make this component usable in different places. And i think it will be easier to maintain & test or transform into a single page.

AnthonySendra: If the logic with props is to complex, if this modal is too different from the original Modal: I'm agree to keep two differents modals

jenow:
+1

PR #22, opened 2016-07-27
KUZBO-116 Enriched user list as per mockup

xbill82:
The list items in the user list section have been enriched according to the [Balsamiq Mockup](https://kuzzleio.mybalsamiq.com/projects/kuzzle-bo-ux/%5Bsecurity%5D%20Users%20%28collapsed%20filters%29) we discussed.

~~:warning: This branch also merges all the commits from @stafyniaksacha #21 in order to have the new multi-profile fixtures.~~

TODO
- [x] Tests

PR #29, opened 2016-07-28
Delete a single user from the user list

xbill82:
:warning: Merge #22 before this!
- [ ] tests (LOL)


jenow:
I am closing this pull request to re implement the delete single user with the bulkDelete action and passing only one id to it. This will avoid writing more tests and harder maintainability for the same action.

PR #30, opened 2016-07-29
Add breadcrumb

stafyniaksacha:


jenow:
+1

src/components/Data/Breadcrumb.vue
```diff
@@ -0,0 +1,129 @@
...
+      <li :class="{'active': $route.params.index}">
+        <a href="#!"
+           v-link="{name: 'DataIndexes'}">
+          <i class="fa fa-globe" aria-hidden="true"></i>
+          Indexes
+        </a>
+      </li>
+
+      <li :class="{'active': $route.params.collection, 'active in': isRouteActive($route, 'DataCreateCollection')}"
+          v-if="$route.params.index">
```
AnthonySendra: I think we should use a `props` given by its parent instead of using `$route`.
We must avoid to use `$route` `$router` in children components

PR #32, opened 2016-08-01
KUZBO-123 create collection

jenow:
Added `create Collection` feature


src/components/Data/Collections/Create.vue
```diff
@@ -1,11 +1,86 @@
 <template>
   <div class="wrapper">
-    CollectionCreate
+    <headline>
+      Create a collection in <i>{{$route.params.index}}</i>
```
AnthonySendra: Could we use props instead of accessing `$route`?
And thanks to vue-router-sync, in parent component, you could use the @stafyniaksacha getter that will provide you the current index in store (= index in url)

src/components/Data/Collections/Create.vue
```diff
@@ -1,11 +1,86 @@
...
+        collectionName: null,
+        error: '',
+        isRealTime: false
+      }
+    },
+    methods: {
+      doCreateCollection () {
+        let mapping = this.$refs.jsoneditor.getJson()
+        this.createCollection(this.$route.params.index, this.collectionName, mapping, this.isRealTime).then(() => {
+          this.$router.go({name: 'DataIndexSummary', params: {index: this.$route.params.index}})
```
AnthonySendra: Same thing here: use props instead of `$route`

src/components/Data/Collections/Create.vue
stafyniaksacha: I think we have to merge this PR before
I will merge the conflits after on the #37 an re check existences of $route in views

src/vuex/modules/data/actions.js
```diff
@@ -28,8 +30,22 @@ export const listIndexesAndCollections = (store) => {
...
+              let realtimeCollections = JSON.parse(localStorage.getItem('realtimeCollections') || '[]')
+              // Cannot use realtimeCollections.find(...) as it is a prototype in ES6 and is not transpiled by babel
+              for (var i = 0; i < realtimeCollections.length; i++) {
+                if (realtimeCollections[i].index === index) {
+                  currentIndex = realtimeCollections[i]
+                  break
+                }
+              }
+              if (currentIndex) {
+                result.realtime = realtimeCollections.map(o => {
```
AnthonySendra: Why don't directly use `realtimeCollections.map`?
Ok, we will loop on "all" realtimeCollections even if there is no realtime collection for this index... but... I think it's clearer and about perf... Array.prototype.map = ~10000000 ops/sec :)

stafyniaksacha: Be carefull about `result.realtime =` it can contain some realtime collection returned by kuzzle itself

PR #36, opened 2016-08-03
Add toaster directive

AnthonySendra:
This directive can't be test because of global variable `Materialize`.

**How to use**
In any component, use `this.$dispatch('toast', text/html, cssClass, duration, cb)`
* text/html is the only required parameter 
* cb is executed when the toast is closed
* "error" is the only current available cssClass

### DIFF
--- a/src/directives/toaster.directive.js
+++ b/src/directives/toaster.directive.js
@@ -1,0 +1,9 @@
+export default {
+  id: 'toaster',
+  bind () {
+    /* eslint no-undef: 0 */
+    this.vm.$on('toast', (text, cssClass = '', duration = 5000, cb) => {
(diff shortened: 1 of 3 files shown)


src/directives/toaster.directive.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  id: 'toaster',
+  bind () {
+    /* eslint no-undef: 0 */
```
stafyniaksacha: Should use `/* eslint-disable-next-line`

`/* eslint ` disable rule for all project

AnthonySendra: Nop, only for the current file

stafyniaksacha:
+1

jenow:
+1

PR #37, opened 2016-08-03
Small refactor: remove $route refs

stafyniaksacha:
Remove `$route` reference in all views.
Instead, forward `index` & `collection` into `props` for all child view's

Added a watcher on `index` property into indexes browse view: when a user change selected index in the tree view, collection list is now refreshed well.

Added also ` --allow-root` for `bower install` due to building in a docker environment (which is done with `root` user)

### DIFF
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -15,8 +15,8 @@
       </li>
       <li>
         <treeview
-          :index="$route.params.index"
-          :collection="$route.params.collection"
+          :index="currentIndex"
+          :collection="selectedCollection"
           :tree="indexesAndCollections">
         </treeview>
       </li>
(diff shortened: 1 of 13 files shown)


jenow:
+1

src/components/Data/Layout.vue
```diff
@@ -43,7 +50,7 @@
 <script>
   import {listIndexesAndCollections} from '../../vuex/modules/data/actions'
   import {getError} from '../../vuex/modules/common/getters'
-  import {indexesAndCollections} from '../../vuex/modules/data/getters'
+  import {indexesAndCollections, selectedIndex, selectedCollection} from '../../vuex/modules/data/getters'
```
AnthonySendra: nitpick: `currentIndex` `currentCollection`?

stafyniaksacha: I have renamed to selected*

PR #44, opened 2016-08-04
Create first admin

AnthonySendra:
:warning: Works only with https://github.com/kuzzleio/kuzzle/pull/367
* Inject bluebird into kuzzle: you can now use `kuzzle.checkTokenPromise`, ...
* Add utility file kuzzleWrapper: this file add functions that not directly exist into kuzzle sdk but are necessary for BO (like `isConnected` that return a promise and reject after 10s)
* Add new component Signup only displayed if there is no user in Kuzzle


jenow:
+1

src/main.js
```diff
@@ -3,15 +3,32 @@ import router from './services/router'
...
+    return Promise.resolve()
+  })
+  .then(() => {
+    router.start({
+      store,
+      components: { App }
+    }, 'body')
+  })
+  .catch((err) => {
+    /* TODO: do something with this... Kuzzle is not responding, what we have to do?? */
```
xbill82: IMHO something is being done wrong here. Roughly, we are executing some logic *before the application starts*, which seems fair. This logic checks whether Kuzzle Server is up or not, whether the user cookie exists or not, whether the token contained in this cookie exists or not, whether admins and users have been properly initialized or not. Fair.
But I think that the app should be able to be alive independently of these factors. IMHO
* the application boots (`router.start` is called)
* some logic checks the availability of Kuzzle Server. If Kuzzle is down, a nice error message is shown. Stop here.
* some logic checks the user cookie, token, firstAdmin, etc... Display relevant loaders and screens in reaction to the state.

Sounds ok?

xbill82:
A part from my remark, :+1:

src/main.js
stafyniaksacha: Yep, good point, but I think we have to discus about this and create a new PR to improve the app bootloading

docker-compose/dev.yml
```diff
@@ -35,7 +35,6 @@ kuzzle:
     - proxy:api
   environment:
     - MQ_BROKER_ENABLED=1
-    - kuzzle_fixtures=/fixtures.json
```
stafyniaksacha: huuuu ?? why have you removed the fixtures ?

stafyniaksacha:
+1

src/main.js
AnthonySendra: Seems good, but we can't do that I think :/
Before `router.start` we have to check everything for redirect the user to the right route, check rights, and check if kuzzle is alive...
But let's talk about this, it's a good subject

docker-compose/dev.yml
AnthonySendra: OOOPS.
Thanks :)

PR #45, opened 2016-08-04
Watch data refactor

stafyniaksacha:
Make the watch data more user friendly
- Removed watch logic from states
- Added alert on too many messages

Tree menu (data) now redirect user relatively (meaning, if the user is on watch data page for a collection, it will be redirected to watch data on the others collections)

Hide browse data in the breadcrumb if the collection is realtime only

Update docker-compose files:
- update ES from 2.2 to 2.3.4
- update Proxy into develop (needed to use subscriptions on latest version of kuzzle)

Added Raleway font
- Set Arial font (alternative) to navigation link, buttons and headings 

Todo: 
- [x] filters watch data
- [x] tests


src/components/Data/Collections/Watch.vue
```diff
@@ -49,51 +129,275 @@
...
+      let toolbar = document.getElementById('notification-controls-fixed')
+
+      this.notifications = []
+
+      window.onscroll = function () {
+        scrolled = true
+      }
+
+      // delay position checking in an interval instead of onscroll event to reduce lags
+      this.scrollListener = setInterval(function () {
```
AnthonySendra: This is the logic for scrollglue directive. We don't have to put this logic in parent component, because if several components need this directive, we have to duplicate code.

src/components/Data/Collections/Watch.vue
```diff
@@ -1,45 +1,125 @@
...
+          <div v-if="!notifications.length" class="inline-alert grey lighten-3">
+            You have not received any notification yet
+          </div>
         </div>
       </div>
     </div>
+
+    <div id="notification-container"
+         v-if="notifications.length"
+         v-scroll-glue:element-tag.body="{items: notifications, active: scrollGlueActive}">
```
AnthonySendra: `v-scroll-glue:element-tag.body`, WAT?

src/assets/global.scss
```diff
@@ -1,4 +1,5 @@
 @import './variables';
+@import './raleway';
```
jenow: I thought we were using Arial instead, can you confirm?

src/components/Data/Collections/Watch.vue
```diff
@@ -49,51 +129,275 @@
...
+      toggleSubscription () {
         if (!this.subscribed) {
           this.subscribed = true
-          this.room = this.subscribe(index, collection)
+          this.room = this.subscribe(this.filters, this.index, this.collection)
         } else {
+          this.reset()
+        }
+      },
+      notificationToMessage (notification) {
```
jenow: Why did you put this in the component? I really think this should be a private utility function.

src/components/Materialize/MSelect.vue
```diff
@@ -6,6 +6,11 @@
   export default {
     id: 'm-select',
     twoWay: true,
+    update (value) {
```
jenow: This is a directive. It shouldn't be a component with an empty template nor in a .vue file

src/assets/global.scss
stafyniaksacha: Nope, Raleway is the main font, and Arial is the secondary one

jenow:
+1

xbill82:
If I shrink a little bit the screen width (like less than 1612px) I start having layout problems on the "Scroll on new messages", "Clear Messages" labels. If I shrink further, the layout of the quick search scrambles up. I'll fix that this afternoon.

src/assets/global.scss
xbill82: Are you sure [Raleway](https://www.google.com/fonts/specimen/Raleway) is the "main" Materialize font? It doesn't look Material Design at all... 
If I go to the Materialize website and inspect any random `<h2>` tag, the font I find is rather [Roboto](https://www.google.com/fonts/specimen/Roboto) which, to me, looks much more Material Design than Raleway.

stafyniaksacha: I Agree with you, I personally prefer Roboto, but it's written in the kuzzle graphic chard that we have to use Raleway ( :cry: )

stafyniaksacha:
Huu yeah good point @xbill82 it's due to the implementation with grid system :(

src/assets/global.scss
xbill82: Ok, now I understand a bit better.

The choice of the font is a design choice.
At the moment, mixing the Kuzzle Design Guidelines with the Material Design Guidelines leads to an ugly result so, IMHO, we should stick with the Material. Currently, the Kuzzle Design Guidelines have been created with print and website in mind, not the BO. To me, it's not the same context.
If we want to conciliate the BO with the Kuzzle Design Guidelines, we should enrich them with this new context (the one of an interaction-intensive application, rather than a mostly read-oriented experience).

Roboto is a great font for the interaction-intensive applications because it is elegant and clear at the same time. It provides great transversal legibility, which is exactly what we need for buttons and labels (which are the main bricks of our UI).
Raleway is great for brand-oriented content, it's elegant but better optimized for blocks of texts rather than buttons and labels. It's not really a "pragmatic" font.

In conclusion, IMHO we should use Roboto, because it is more coherent with Material and better for the user.

PR #46, opened 2016-08-05
Crudl documents component

jenow:
Added CrudlDocument component and integrated it in security/user


src/components/Common/CrudlDocument.vue
```diff
@@ -0,0 +1,240 @@
...
+      Filters
+    },
+    props: [
+      'index',
+      'collection',
+      'documents',
+      'displayBulkDelete',
+      'allChecked',
+      'lengthDocument'
+    ],
```
stafyniaksacha: can you describe the props ? with type and default values

src/components/Common/CrudlDocument.vue
```diff
@@ -0,0 +1,240 @@
...
+      refreshSearch () {
+        this.$router.go({query: {...this.$route.query, from: 0}})
+      },
+      dispatchToggle () {
+        this.$dispatch('toggle-foo')
+      }
+    },
+    events: {
+      'perform-search' () {
+        console.log('searchTerm', this.searchTerm)
```
stafyniaksacha: console.log

src/components/Security/Users/List.vue
```diff
@@ -166,72 +70,11 @@
...
-        }
-
-        if (this.sorting) {
-          sorting = formatSort(this.sorting)
-        }
-
-        // Execute search with corresponding filters
-        this.performSearch('users', '%kuzzle', filters, pagination, sorting)
+    events: {
+      'toggle-foo' () {
```
stafyniaksacha: what is `toggle-foo` ?

stafyniaksacha:
+

src/vuex/modules/common/crudlDocument/store.js
```diff
@@ -24,6 +21,16 @@ export const mutations = {
     state.documents = result.documents
     state.total = result.total
   },
+  [TOGGLE_SELECT_DOCUMENT] (state, id) {
```
AnthonySendra: I thought we had said that we no used store for selected document?

jenow: I know but since the documents are fetched on the ready action of the component it is not a problem for them to be in the store (no global problem caused). If this is a problem for you then we should really discuss about when to put things in the vuex store.

PR #52, opened 2016-08-09
Login UX + Change default route "/" to index/browse component

stafyniaksacha:


xbill82:
The PR is LGTM, but I think we should provide a better error reporting on login failure. Currently, users won't know whether the login failed because Kuzzle is broken or because the provided credentials are bad.

But this should be a separate ticket.

stafyniaksacha:
Yep, I agree with your remark, but I think we have to dig into an error reporting system separately

src/components/Login.vue
```diff
@@ -1,42 +1,66 @@
...
-          <div class="row">
-            <div class="col m6">
-              <p class="error">{{error}}</p>
-            </div>
-            <div class="col m6">
-              <p class="right-align">
-                <button class="btn waves-effect waves-light" type="submit" name="action">Login</button>
-              </p>
-            </div>
+            <form class="col s10 offset-s1" id="loginForm" method="post" @submit.prevent="login(username, password)">
```
AnthonySendra: Why do we send `username` and `password` in parameters? We have directly access in method with `this.username` thanks to `v-model`

stafyniaksacha: ok for feedback; will be implemented into the #55

AnthonySendra:
:+1:

jenow:
+1

PR #60, opened 2016-08-15
add global loader + route loader

stafyniaksacha:
Added UX for low latency network
- Pre loader *(in `index.html` ~ 2.3kb once builded)* which display a loader until the application is ready to start
- Smooth loader while lazy loading new routes

Steps to reproduce: 
- go to the backoffice on chrome
- open chrome web developer
- go to network tab > switch the throttling selector to "Regular 3G ..."
- refresh the backoffice
- navigate through pages

*there is a css style minified in the `index.html` file, it's because webpack doesn't compile inline styles on `npm run build`*

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,9 +3,17 @@
   <head>
     <meta charset="utf-8">
     <title>kuzzle-backoffice</title>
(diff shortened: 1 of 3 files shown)


jenow:
+1

PR #63, opened 2016-08-18
Kuzzle is disconnected

AnthonySendra:
* Implement special page when the BO can't connect to Kuzzle
* Implement modal with error message


src/components/Error/KuzzleDisconnectedPage.vue
```diff
@@ -0,0 +1,61 @@
...
+
+</style>
+
+<script>
+  import kuzzle from '../../services/kuzzle'
+  import KuzzleDisconnected from './KuzzleDisconnected'
+  import { setConnection } from '../../vuex/modules/common/kuzzle/actions'
+
+  let idConnect
+  let idReconnect
```
jenow: Shouldn't this be in data ?

AnthonySendra: Nop, we don't have to bind something, nor to display these variables

jenow:
+1

stafyniaksacha:
+1

PR #71, opened 2016-08-23
Collection list refactoring

xbill82:
Fixes #66 

Various architecture and test refactoring have been performed.

* The `store.data.collections` object has been demolished :bomb: 
* The computation of the number of collections inside an index has been moved to a service.
* The extraction of the collections from the index tree has been placed into a service.
* Tests have been updated.
* Components do not contain much logic, In Nomine Patris, Dan Abramov et Spiritu Sancti, Amen.

PR #72, opened 2016-08-23
Deduped realtime collections that are also stored.

### LINKED ISSUE #67: Duplicate collection in treeview storage/realtime
How to reproduce?

* go to http://localhost:3000/#!/data/kuzzle-bo-testindex/editable-collection/watch
* click on subscribe
* open an other window on  http://localhost:3000/#!/data/kuzzle-bo-testindex/

Collection `editable-collection` is now in storage and in realtime. We have to exclude from realtime collections already in present in storage.

xbill82:
Fixes #67

### DIFF
--- a/src/vuex/modules/data/actions.js
+++ b/src/vuex/modules/data/actions.js
@@ -1,4 +1,5 @@
 import kuzzle from '../../../services/kuzzle'
+import { dedupeRealtimeCollections } from '../../../services/data'
 import Promise from 'bluebird'
 import {
   RECEIVE_MAPPING,
@@ -45,6 +46,7 @@ export const listIndexesAndCollections = (store) => {
             }
             if (index !== '%kuzzle') {
               addLocalRealtimeCollections(result, index)
+              result = dedupeRealtimeCollections(result)
 
               indexesAndCollections.push({
                 name: index,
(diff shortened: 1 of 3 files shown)


AnthonySendra:
:+1:

src/vuex/modules/data/actions.js
```diff
@@ -45,6 +46,7 @@ export const listIndexesAndCollections = (store) => {
             }
             if (index !== '%kuzzle') {
               addLocalRealtimeCollections(result, index)
+              result = dedupeRealtimeCollections(result)
```
stafyniaksacha: Hum, 2 point here: 
- `dedupeRealtimeCollection()` should be called after each call of `addLocalRealtimeCollections()` -> so you've missed a [call at line 87](https://github.com/kuzzleio/kuzzle-backoffice/blob/627a7ae802858e696d764d93203c52c385b5ea1a/src/vuex/modules/data/actions.js#L87)
- OR `dedupeRealtimeCollection()` logic should be added into `addLocalRealtimeCollections()`

What did you think ?

stafyniaksacha: Ok, didn't see that you have removed this part in #71 !

stafyniaksacha:
+1

PR #103, opened 2016-09-13
Clean unit tests

AnthonySendra:
Remove all warning messages from Vue in unit tests


jenow:
+1 well done

src/vuex/modules/data/actions.js
```diff
@@ -63,7 +63,7 @@ export const listIndexesAndCollections = (store) => {
           store.dispatch(RECEIVE_INDEXES_COLLECTIONS, res[0])
           resolve()
         })
-        .catch((error) => reject(new Error(error.message)))
+        .catch((error) => reject(error))
```
AnthonySendra: Because it's already an error :D

stafyniaksacha: Yep, but we always need to recreate the error object to have the real error stack trace

PR #131, opened 2016-09-21
Disconnected state refactoring

xbill82:
This PR introduces a (medium sized) refactoring of the "disconnected from Kuzzle" state of the application.
This is necessary to be able to switch between environments.

So far, we've been managing this state at bootstrap (in `src/main.js`), but now the application can be disconnected at any time (whenever we switch from an environment to another). So all the "init" operations need to be now performed by a specific service.

Some of the major changes introduced in this PR:

* New `environments` service, allowing to CRUD the known environments.
* The `kuzzle` service now returns a disconnected instance of the SDK. The application must explicitly connect it to an environment (`host` + `wsPort`).
* The `KuzzleDisconnected` page is now handled by `App` instead of the router, since it's more like a macro-state of the application, rather than a route that the user would be eventually requesting. While disconnected, only this page is shown as the content of `App` (instead of the classic `router-view`).
* The application can now be started via the `bin/start.js` script
* All the dependencies have been moved to `devDependencies` except for the ones requested by `bin/start.js`

This is a WIP pull-request because I need some feedback about the refactoring. The tests will be written once the refactoring is approved.


package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "node build/dev-server.js",
     "build": "node build/build.js",
     "prod": "npm run build && npm start",
-    "start": "./node_modules/.bin/http-server -p ${BACKOFFICE_PORT:=3000} dist/",
+    "start": "./node_modules/.bin/start --port ${BACKOFFICE_PORT:=3000}",
```
AnthonySendra: Because you added `rc`, for overriding the port, I think the environment variable is something like `KUZBO_PORT` https://www.npmjs.com/package/rc#standards

bin/start
```diff
@@ -0,0 +1,18 @@
...
+var app = express()
+
+app.use(express.static(path.resolve(__dirname, '../dist')))
+
+module.exports = app.listen(rcConf.port, function (err) {
+  if (err) {
+    console.log(err)
+    return
+  }
+  console.log('Listening at http://localhost:' + rcConf.port + '\n')
```
scottinet: Shouldn't this be `` `Listening at http://${rcConf.host}: ${rcConf.port}\n` ``?

AnthonySendra:
At this point, LGTM

package.json
xbill82: yes exactly, you're right

xbill82: Well... I checked, it's actually `KUZBO_port`, which is ugly (case mess-up). I'd like to set the app name to `kuzbo` instead. Is it ok to have env vars in lowercase?

src/components/Error/KuzzleDisconnectedPage.vue
```diff
@@ -1,11 +1,24 @@
...
-    <div class="row">
-      <div class="col s12">
-        <kuzzle-disconnected :host="host" :port="port"></kuzzle-disconnected>
+  <div class="container-error">
+    <div class="container">
+      <div class="row">
+        <div class="col card wrapper s6 offset-s3">
+          <div class="col s12">
+            <h1 class="center-align logo">
+              <img src="../../assets/logo.png" alt="Kuzzle Backoffice" style="width: 70%"/>
```
jenow: style? What year is it!

jenow:
+1

xbill82:
Ok, just fixed the failing tests, this afternoon I'll fix the coverage and all will be ready and set to implement the UI layer.

package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "node build/dev-server.js",
     "build": "node build/build.js",
     "prod": "npm run build && npm start",
-    "start": "./node_modules/.bin/http-server -p ${BACKOFFICE_PORT:=3000} dist/",
+    "start": "./node_modules/.bin/start",
```
stafyniaksacha: it's `./bin/start` here no ?

xbill82: oh :octopus:

AnthonySendra:
You have some warnings in your tests

```
Home.vue tests
    watch tests
      tokenValid
        ✓ should open the login modal if value pass to false
        ✓ should do nothing if value pass to true
      kuzzleIsConnected
        ✓ should close the modal if kuzzle is connected
        ✓ should open the modal if kuzzle is connected
    methods tests
      onLogin
        ✓ should close the login modal if value pass to false
WARN LOG: '[vuex] Action bound to key 'vuex.actions.doLogin' is not a function.'
  LoginForm.vue
    methods Tests
      login
        ✓ should display the error when login fail
WARN LOG: '[vuex] Action bound to key 'vuex.actions.doLogin' is not a function.'
        ✓ should call onLogin callback if success
```

```
loginFromCookie action
    ✓ should login user from cookie
    ✓ should not login user from cookie because the jwt token is wrong
    ✓ should resolve null and do nothing if there is no user
    ✓ should resolve null and do nothing if there is user in cookie but with bad token
WARN LOG: 'Warning: a promise was rejected with a non-error: [object Error]'
WARN LOG: 'Unhandled rejection Error: undefined is not a function (evaluating '_kuzzle2.default.unsetJwtToken()')'
  checkFirstAdmin action
    ✓ should reject if error comes from Kuzzle
    ✓ should dispatch true if admin already exists
    ✓ should dispatch false if there is no admin
```

```
 Signup component
    Methods
      signup
        ✓ should set error if one of required fields is empty
        ✓ should set error if password does not match
        ✓ should set waiting if everything is fine
ERROR LOG: Error{}
        ✓ should redirect on Login if there is an error
        ✓ should call setFirstAdmin and redirect to Login if everything is ok
        ✓ should call queryPromise with right username/password and reset false
        ✓ should call queryPromise with right username/password and reset true
```

xbill82:
@AnthonySendra done! We just left the `ERROR LOG`, which is simply `console.log()`ged  in the code (so we added a TODO to remind us to propagate the error message back to the UI eventually)

PR #133, opened 2016-09-27
Persist environments in localStorage

xbill82:
With this PR, we enable the back office to store the user's environments in the browser's localStorage.

This PR is set as WIP because I need to know if I'm lowering the coverage or not.

Solves #128


src/services/environment.js
```diff
@@ -2,33 +2,140 @@ import {
...
+
+  let newEnvironment = {
+    name,
+    color,
+    host,
+    ioPort,
+    wsPort
+  }
+
+  kuzzleActions.addEnvironment(store, generateHash(name), newEnvironment)
```
jenow: I don't understand why you hash the name of the environment

stafyniaksacha: :question: same to me

xbill82: oh this is totally pointless.
Planned to wipe it out.

PR #213, opened 2017-02-09
Use the same port for both websocket and socketio protocols 

ballinette:
Following 
https://github.com/kuzzleio/kuzzle-proxy/pull/48
https://github.com/kuzzleio/sdk-javascript/pull/180

Use now the same port (`7512`) for both websocket and socketio protocols 

TODO: update the JS SDK version in `package.json` after the deployment of https://github.com/kuzzleio/sdk-javascript/pull/180

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "karma-sourcemap-loader": "0.3.7",
     "karma-spec-reporter": "0.0.26",
     "karma-webpack": "1.8.0",
-    "kuzzle-sdk": "3.1.1",
+    "kuzzle-sdk": "3.2.0",
     "lolex": "1.5.1",
     "materialize-css": "0.97.6",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-backoffice",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
AnthonySendra: I think it will be 1.2.0

ballinette: I guess we just have a prerelease right now with version 1.1.1, and then include the changes you are working on to the version 1.2.0 ?

AnthonySendra: In fact, that will be the version 2.0.0: https://github.com/kuzzleio/kuzzle-backoffice/milestone/4?closed=1

Issue #207, opened 2017-01-24
Switch environment on create


AnthonySendra:
When the user create a new environment on the login page or on the error page we must redirect him to the new environment

AnthonySendra:
Fixed on 2.1.0

PR #236, opened 2017-03-27
Create environment dedicated page when no lastConnectedEnv

AnthonySendra:
* [x] Create the dedicated create environment page
* [x] Remove default environment on localhost
* [x] Check if lastConnectedEnv is defined and redirect on create env page if not
* [x] Automatic switch on newly created environment


AnthonySendra:
Fix #236 and #207

PR #272, opened 2017-05-30
Fix publication on NPM

xbill82:
# Description

The `dist` folder was included in the VCS to ensure the published npm package would include the built version. This polluted the git status output and the PR review.
This was due to npm [falling back to `.gitignore` for exclusion](https://docs.npmjs.com/cli/publish) when `.npmignore` is not present.

This PR introduces a valid `.npmignore`, excludes `dist` from the VCS and adds the `beforepublishOnly` (which is a transitional hook, that [will be dropped in npm@6.0.0](https://docs.npmjs.com/misc/scripts) :warning: ) to trigger the build before publishing.

# Boyscouting

Re-introduced the NPM `start` script, which enables `kuzzle-backoffice start` as per the README file.


xbill82:
@scottinet oh... How should we proceed then?

scottinet:
No sweat: once this PR is merged just tag the team so that we can take the necessary actions. :)

xbill82:
Merging as 2 :white_check_mark:

PR #276, opened 2017-05-31
Prevent a document from being created more than once

scottinet:
# Description

Disables the `Create` button in the document creation form once it has been submitted to prevent a document from being created more than once

# Solved issue

#269


xbill82:
@benoitvidis Do you refer to pressing Enter from within a form field? In this case, yes: the "submit on enter" behavior should be inhibited once the state switches to `submitted`.
If you rather refer to pressing enter on the button (like, focusing the button and pressing enter), I doubt that the button triggers anything when disabled.

scottinet:
@xbill82 > I was unable to reproduce the problem using the `enter` key with the fix enable, but @benoitvidis and I discovered why: modern browsers attach the submit action to the first `submit` component it finds. So disabling the button effectively disable any further form submits.
But this only works on some browsers. So I've already made a small change following Benoit's remark to disable any further document creation once one is underway.

xbill82:
@benoitvidis can we merge this?

PR #277, opened 2017-06-02
Add a truncate collection dropdown option

samniisan:
# Introduces
- Dropdown option in Collection view, with confirmation modal, to truncate a collection


src/vuex/modules/collection/actions.js
```diff
@@ -93,5 +93,19 @@ export default {
 
     localStorage.setItem('defaultJsonView', JSON.stringify(indexes))
     return commit(types.SET_COLLECTION_DEFAULT_VIEW_JSON, {jsonView})
+  },
+  [types.CLEAR_COLLECTION] ({state}, {index, collection}) {
```
jenow: You should unit test this

samniisan: @jenow Done

samniisan:
@scottinet
- Tested on various browsers, the list will be empties once PR https://github.com/kuzzleio/kuzzle/issues/854 is merged on core side
- There are currently no "success" toast on BO as a design UX choice, checked with @jenow . As long as there is enough visual feedback that the action performed the awaited result (here, the emptied collection), there's no need to add a notification.

scottinet:
Ok :+1:

Issue #273, opened 2017-05-30
Form to edit documents & security items


xbill82:
@dbengsch commented on [Fri Mar 10 2017](https://github.com/kuzzleio/kuzzle-private-backlog/issues/19)

---

@gandbox commented on [Wed Mar 22 2017](https://github.com/kuzzleio/kuzzle-private-backlog/issues/19#issuecomment-288398028)

Missing : security forms (users, policies, roles)

PR #282, opened 2017-06-07
User creation and update via user-friendly wizards

xbill82:
# Related issues
#273 
#246

# Description

Instead of simply providing a JSON editor for users, we now provide the user with a 3-step wizard that mimics the Collection mapping and Document creation / update process. The user can now

* Set the mapping for the "Custom User Data" (i.e. the data stored in the `content` field of each user object, in the `%kuzzle` index) - This data represents a payload attached to every user, regardless of the strategies they are authenticated to.
* Create a user via a 3-step wizard, that progressively shows the "Basic", "Credentials" and "Custom" forms for the user. The "Custom" view can be switched between Form and JSON (the Form is automatically generated based on the mapping defined above).
* Update a user via a 3-tabs panel (the tabs are not steps, the user can freely navigate through them).

# Internals

Some components have been generalized from the Collections / Documents sections, since they were relying on the data in the Store. They are now functional, meaning that they are context-agnostic: they receive the data via the props and pop it via the events. The following components have been generalized

* The JsonForm and JsonLine, i.e. the ones that allow building a form based on an existing mapping and a schema (that maps the mapping fields with a set of HTML elements).
* The Stepper.
* The JsonEditor with Mapping.

These changes have not been applied to the existing work made on the Collections and Documents. I'll try to make a PR for this as soon as I have the time.

# Notes

Improvements on the wording are welcome.


src/components/Common/JsonEditor.vue
```diff
@@ -39,13 +45,14 @@
         if (this.height === undefined) {
           return {height: '100px'}
         } else {
-          return {height: this.height}
+          return {height: this.height + 'px!important'}
```
stafyniaksacha: missing space: `px!important` -> `px !important`

PR #297, opened 2017-06-13
Wrap notification messages in an auto scroll div

### LINKED ISSUE #286: Display bug on collection's watch tab
On collection's watch tab, when the main page display starts to have a scroll bar, it is not possible to interact with both filter builder and published messages / documents; the arrows won't unfold, and when unfolded, will fold as soon as we try to scroll.

jenow:
Wrap notification messages in an auto scroll div so the body won't scroll anymore.
fix #286


src/components/Data/Collections/Watch.vue
```diff
@@ -160,12 +158,9 @@
 </style>
 
 <script>
-  import {debounce} from 'lodash'
   import CollectionTabs from './Tabs'
   import Headline from '../../Materialize/Headline'
 
-  import ScrollGlue from '../../../directives/scroll-glue.directive'
```
stafyniaksacha: You can remove this directive if it is not used anymore

src/components/Data/Collections/Watch.vue
```diff
@@ -370,6 +365,12 @@
         }
 
         this.notifications.push(this.notificationToMessage(result))
+
+        // Auto scroll
+        const div = document.getElementById('notification-container')
```
stafyniaksacha: here you can use refs instead of `getElementById`

jenow: I used it on DOM element of the current component, you can't access a ref from a child element

src/components/Data/Collections/Watch.vue
jenow: Done !

PR #339, opened 2017-06-29
 Fixed KuzzleErrorPage

### LINKED ISSUE #329: Stuck on "loading" view
When starting the BO with no kuzzle instance up, we don't get the "Kuzzle unavailable" view (allowing to create and switch environments), but we're stuck on the loading page instead.

selection_033

xbill82:
Fixes #329 

An error prevented the KuzzleErrorPage to be displayed, which resulted in being stuck on the "Loading" page. The error is fixed.

## Boyscouting

Previously, if the BO page was re-loaded during Kuzzle downtime (i.e. while the KuzzleErrorPage was displayed), the environment initialization process would result in an error (which is good), leaving the `adminAlreadyExists` variable to a falsey value. Once Kuzzle went up again, the environment initialization was not resumed. Simply, the `errorFromKuzzle` attribute was set to `null`. This would set the conditions for the Signup page to show up, even in cases it was not necessary.
Now, the BO re-tries to connect to the last environment as soon as Kuzzle is resumed.

### DIFF
--- a/src/components/Error/KuzzleErrorPage.vue
+++ b/src/components/Error/KuzzleErrorPage.vue
@@ -11,14 +11,14 @@
     <div class="row message-warning">
       <h5>{{$store.state.kuzzle.errorFromKuzzle.message}}</h5>
       <div class="divider"></div>
-      <div class="message" v-if="errorStatus">
+      <div class="message" v-if="errorInternalStatus || errorInternalMessage">
         <i class="fa fa-plug"></i>
-        [{{ errorStatus }}]
+        [{{ errorInternalStatus }}] {{ errorInternalMessage }}
       </div>
     </div>
     <div class="row kuzzle-disconnected">
       <div class="col s12">
-        <p>I'm doing my best to reconnect to Kuzzle ...</p>
+        <p>Trying to connect to Kuzzle...</p>
       </div>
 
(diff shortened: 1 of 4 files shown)


src/components/Error/KuzzleErrorPage.vue
```diff
@@ -11,9 +11,9 @@
     <div class="row message-warning">
       <h5>{{$store.state.kuzzle.errorFromKuzzle.message}}</h5>
       <div class="divider"></div>
-      <div class="message">
+      <div class="message" v-if="errorStatus">
         <i class="fa fa-plug"></i>
-        [{{$store.state.kuzzle.errorFromKuzzle.internal.status}}] {{$store.state.kuzzle.errorFromKuzzle.internal.message}}
+        [{{ errorStatus }}]
```
stafyniaksacha: Why did you discard internal message error ? It's important for end user

xbill82: I found that it was redundant with the `$store.state.kuzzle.errorFromKuzzle.message`. My stand is that we should keep just one, but the choice is arbitrary.
Also, I would suggest to drop the `internal.status` code, which IMHO is useless for the user. What do you think?

PR #340, opened 2017-06-30
Create/Update user UX

### LINKED ISSUE #314: Improve credentials step of user creation
Can we think about a better presentation for fields? The right part is not clear.
Plus, the view is totally different between create/update: the drop down is not present on update.

AnthonySendra:
fix #314 

* [x] Use a form view in credential instead of JSON editor
* [x] Refactor for using async/await
* [x] Avoiding multi-call on the same route by moving the call in the corresponding component (the credential component will call Kuzzle for getting the `credentialMapping`, the custom component will call Kuzzle for gettting the `customContentMapping`, ...)
* [x] Split component and add a new StepsContent that contains basic/credentials/custom
* [x] Use v-model where we can
* [x] Implement functional testing in Testim


stafyniaksacha:
On `/security/users/create` 

* step 1:
When checking `Auto-generate` option on KUID field an error occurs saying `Please fill the custom KUID or check the auto-generate box`
* step 3:
We have editable `_kuzzle_info` field 

Note about `_kuzzle_info` field 
If we keep it empty on creation step, and then edit the user document, we see that theses metadata are not automatically updated

ballinette:
I've the same error as Sacha, but with following message: `Security.createUser: cannot create a user without a user ID`

ballinette:
Ok, it was a bug in the SDK, which has been fixed in 5.0.6 (see https://github.com/kuzzleio/sdk-javascript/pull/229 )

PR #348, opened 2017-07-24
Some unit tests to increase coverage

xbill82:
# Dependencies
Depends on #347, do not merge before it.


src/services/kuzzleWrapper.js
```diff
@@ -267,7 +275,7 @@ export const performSearchRoles = (controllers = {}, pagination = {}) => {
       let roles = result.roles.map((document) => {
         let object = {
           content: document.content,
-          meta: document.meta ? new Meta(document.meta) : null,
+          meta: document.meta ? new Meta(document.meta) : new Meta({}),
```
dbengsch: could also be `meta: new Meta(document.meta || {})`

PR #352, opened 2017-08-30
Fixed profile creation

xbill82:
When creating a Profile, the component [sent the data as an `Object`](https://github.com/kuzzleio/kuzzle-backoffice/blob/2.x/src/components/Data/Documents/Common/CreateOrUpdate.vue#L133), which made Kuzzle complain about the type of the `policies` attribute of the `input.body` of the Request.

selection_070

### DIFF
--- a/src/components/Security/Profiles/Create.vue
+++ b/src/components/Security/Profiles/Create.vue
@@ -49,7 +49,7 @@
 
         kuzzle
           .security
-          .createProfilePromise(this.id, profile, {replaceIfExist: true})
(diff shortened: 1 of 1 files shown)


AnthonySendra:
Shouldn't we also remove the `_kuzzle_info` from the mapping?
What do you think, let's make a ticket?

ballinette:
@AnthonySendra the `_kuzzle_info` has to be removed a Kuzzle side (like for document's mappings).
It's fixed by this PR: https://github.com/kuzzleio/kuzzle/pull/946

PR #361, opened 2017-09-28
Fix collection dropdown not activating

samniisan:
Issue happened when a collection was named with one or more underscores

### DIFF
--- a/src/components/Materialize/Dropdown.vue
+++ b/src/components/Materialize/Dropdown.vue
@@ -55,7 +55,7 @@
 
         let parsed = this.id + this._uid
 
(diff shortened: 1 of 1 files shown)


src/components/Materialize/Dropdown.vue
```diff
@@ -55,7 +55,7 @@
 
         let parsed = this.id + this._uid
 
-        return parsed.replace(/[!"#$%&'()*_+,./:;<=>?@[\]^`{|}~ ]/g, '\\$&')
+        return parsed.replace(/[!"#$%&'()*+,./:;<=>?@[\]^`{|}~ ]/g, '\\$&')
```
scottinet: Collections can also be created with the following characters in their names: `!$%&()*+./:;<=>?@[]^`{}|~` and spaces:

screenshot from 2017-09-28 11-24-31

PR #374, opened 2018-03-12
fix #366 Call logout action when json web token error

### LINKED ISSUE #366: Switching from remote Kuzzle to a kuzzle on localhost hangs
Steps to reproduce:
-Have a Kuzzle running on localhost
-Have a Kuzzle running somewhere in the cloud (e.g. iot.uat.kuzzle.io)
-open backoffice on localhost kuzzle and login
-switch to remote kuzzle (login if needed)
-switch back to localhost, backoffice stucks on 'Trying to connect to Kuzzle' screen
screenshot from 2018-01-08 14-45-30

-pressing F5 to refresh page allow browsing kuzzle again.

-switch back to remote kuzzle works well...

Njuelle:
Fix #366  : dispatch DO_LOGOUT action when "json web token error" when switching environment


scottinet:
As discussed together, we shouldn't be calling `logout` while knowing that the token is invalid, since Kuzzle will simply return a `Invalid JWT` error. Plus, this may trigger a dispatch to logout again, making the code looping endlessly.

Also, can you please check that the correct token is retrieved when switching environment? I don't know if the problem detailed by #366 arises because the token has expired, or because the admin console sends the previous environment token, or because of https://github.com/kuzzleio/kuzzle/issues/1068

PR #378, opened 2018-03-19
fix #362 dispatch LIST_INDEXES_AND_COLLECTION in mounted Treeview

### LINKED ISSUE #362: Backoffice won't load the first time, need to F5 to get the data
Title says it all.

After login, the backoffice show this : 
download 2

I have to refresh to be able to see the index/collections

Njuelle:
Remove dispatch action in mounted Layout and add same action in mounted Treeview


scottinet:
Can you help me understand what the problem was and how does this PR solve #362?

PR #379, opened 2018-03-20
Add lastNotification component when watching collection

Njuelle:
Add lastNotification component : allows to see the last document informations when watching a collection in real time


Njuelle:
capture du 2018-03-21 15-01-44

PR #388, opened 2018-03-27
Fix build error with vue-native-notification

Njuelle:
- Remove dependencie vue-native-notification
 - Manage notifications with native API

### DIFF
--- a/src/components/Data/Collections/Watch.vue
+++ b/src/components/Data/Collections/Watch.vue
@@ -260,7 +260,11 @@
       },
       toggleSubscription () {
         if (!this.subscribed) {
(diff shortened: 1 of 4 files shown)


src/components/Data/Collections/Watch.vue
```diff
@@ -446,11 +446,12 @@
       setScrollDown (v) {
         this.scrollDown = v
       },
-      webNotification (text) {
-        this.$notification.show('Kuzzle Admin Console', {
+      handleWebNotification (text) {
+        let notif = new window.Notification('Kuzzle Admin Console', {
```
scottinet: let => const

PR #391, opened 2018-03-29
Manage index/collection routes exist

Njuelle:
- Add check if index/collection routes exist in Layout component
 - Clean 404 component template

### DIFF
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -40,11 +40,10 @@
     methods: {
       setRouteExist () {
         this.routeExist = true
-        let index = this.$store.state.route.params.index
+        const {index, collection} = this.$store.state.route.params
         if (typeof index !== 'undefined' && this.$store.state.index.indexes.indexOf(index) === -1) {
           this.routeExist = false
         } else {
-          let collection = this.$store.state.route.params.collection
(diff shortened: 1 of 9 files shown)


src/components/Data/Layout.vue
```diff
@@ -21,17 +23,39 @@
...
+    },
+    data () {
+      return {
+        routeExist: true
+      }
+    },
+    methods: {
+      setRouteExist () {
+        this.routeExist = true
+        let index = this.$store.state.route.params.index
```
scottinet: prefer `const` over `let` for constants

src/components/Data/Layout.vue
```diff
@@ -21,17 +23,39 @@
...
+        if (typeof index !== 'undefined' && this.$store.state.index.indexes.indexOf(index) === -1) {
+          this.routeExist = false
+        } else {
+          let collection = this.$store.state.route.params.collection
```
scottinet: prefer `const` over `let` for constants

scottinet: You may also use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to initialize both `index` and `collection` in one command, like so:

`const {index, collection} = this.$store.state.route.params;`

PR #396, opened 2018-04-06
Fix json-formatter 

Njuelle:
Force regenerate html in json formatter directives when update and append it


Njuelle:
Adding `--unsafe-perm` to npm install_deps script to fix phantomJS NPM failed download

PR #414, opened 2018-06-19
corrected grammar (ENG)

appacea:
## What does this PR do ?
Grammar corrections in English

Read changes


bower.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "kuzzle-backoffice",
+  "name": "kuzzle admin console",
   "description": "A handy administrative console for Kuzzle.io",
```
etrousset: Why keep Kuzzle.io here, if we change it later?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "kuzzle-backoffice",
+  "name": "kuzzle-admin-console",
   "version": "2.1.5",
-  "description": "A handy administrative console for Kuzzle.io",
+  "description": "A handy administrative console for Kuzzle",
   "author": "The Kuzzle team <support@kuzzle.io>",
   "private": false,
   "bin": "./bin/kuzzle-backoffice",
```
etrousset: We should not forget to rename the binary also

appacea: yep. didn't want to touch the program part but I can do it if there's no impact

bower.json
appacea: good catch. ill change it.

Issue #408, opened 2018-05-31
Sometimes it's impossible to scroll


Tyki:
On the first loading of the backoffice, **_randomly_** it's impossible to scroll with a lot of indexes

## Expected Behavior

Be able to scroll

## Current Behavior

Impossible to scroll

## Possible Solution

Do a F5 and pray that the console is correctly initialized

## Steps to Reproduce

1. Create some indexes : `for i in {1..40}; do; curl -H "Content-Type: application/json" -X POST  "http://localhost:7512/test$i/_create"; sleep 0.1; done`
2. Login as anonymous to the admin console

## Context (Environment)

Using http://kuzzle-backoffice.netlify.com/#/data

On the first loading, the scrollbar is missing, making it impossible to scroll
![image](https://user-images.githubusercontent.com/2605688/40770167-31a753ca-64ba-11e8-9bed-350edf29cec4.png)

After a clear-cache & reload, the scrollbar is there
![image](https://user-images.githubusercontent.com/2605688/40770194-4287e9c0-64ba-11e8-9f34-9c9db73661ec.png)

Njuelle:
Hi!
I can't reproduce this bug on chrome and firefox.
After you cleared the cache, are you able to reproduce this bug?
If so, what are the steps to reproduce?

Thx

Tyki:
Well @Aschen was able to reproduce the issue when i wrote this ticket

Sometimes, I still have this issue. Not sure how to reproduce

Aschen:
I will show you @Njuelle

PR #426, opened 2018-08-08
408 fix scroll after login

Aschen:
## What does this PR do ?

After login, the `body` has the style `overflow: hidden` and it was impossible to scroll over large number of index.  

Fix: #408 

  - Step 1: create a lot of index `for i in {1..40}; do; curl -X POST "http://localhost:7512/test$i/_create"; sleep 0.1; done`
  - Step 2: Open the admin console and then logout
  - Step 3: Log as anonymous, you should be able to scroll over the index list

### DIFF
--- a/src/components/Login.vue
+++ b/src/components/Login.vue
@@ -61,6 +61,8 @@
     },
     methods: {
       onLogin () {
+        window.document.body.style.overflow = 'visible'
+        
         if (this.$store.getters.routeBeforeRedirect) {
           this.$router.push({name: this.$store.getters.routeBeforeRedirect})
         } else {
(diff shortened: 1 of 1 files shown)


src/components/Login.vue
```diff
@@ -61,6 +61,8 @@
     },
     methods: {
       onLogin () {
+        window.document.body.style.overflow = 'visible'
```
jenow: It would be better to do CSS in CSS file and not through javascript

Aschen: I totally agree but the overflow on the body is [manually set](https://github.com/kuzzleio/kuzzle-admin-console/blob/master/src/components/Materialize/Modal.vue#L60) to `hidden` when the modal is open.  
After login we push the index route in the router and the body overflow is never set to `visible` again.

Njuelle: Can you at least put a comment explaining why we're changing this css property here?

PR #428, opened 2018-08-09
KZL-30 Persistent search filters across collections and reloads

xbill82:
## What does this PR do ?
This PR enables the Data -> Index -> Collection -> Browse view to remember search filters across collections, reloads and document edits.

#### Persistent filters across collections

1. Perform a search on a collection (quick, basic o raw) - you should see the filter specified in the URL
2. Switch to another collection
3. Switch back to the previous collection
4. The filter specified for the previous search is still defined

#### Persistent filters across reloads

1. Perform a search on a collection (quick, basic o raw)
2. The the filter is specified in the URL
3. Reload the page
4. The filter specified for the previous search is still defined

#### Persistent filters across document edits

1. Perform a search on a collection (quick, basic o raw)
2. Edit a document in the results
3. After updating the document, the results of the search are still visible

#### Persistent filters across sessions

1. Perform a search on a collection (quick, basic o raw)
2. Logout and login again
3. Browse to the previous collection
4. The filter specified for the previous search is still defined

Note that this happens even if you login as another user.

### Boyscout

A major refactoring has been performed on the search filters system.

PR #430, opened 2018-08-13
Filters css refactor

xbill82:
~Depends on #428~

## What does this PR do ?
This PR aligns the styles of the components touched in #428 with the guidelines specified in `CONTRIBUTING.md`. Specifically, 

* styles have been moved from global `.scss` files to individual component's scoped `style` tag
* CSS classes have been harmonized with the BEM convention
* many useless style rules have been wiped away

You should just browse data and use the filters and ensure everything looks fine.


Aschen:
I have another bug but it was introduced by the previous PR..
 - Select BasicFilter mode
 - Type something only in the `Attribute` field
 - Submit the form
 - You should see an error and the component is re-render so it switch on the QuickFilter mode
image

PR #432, opened 2018-08-20
KZL 301 - Display realtime search result when updating filters

Aschen:
## What does this PR do ?

~Depends on https://github.com/kuzzleio/kuzzle-admin-console/pull/430 (PR is based on this branch for now, so you can review it)~

This PR refresh the search results in realtime when the query is updated.  
When you are typing in the quick filter input or in the basic filter inputs, requests are made to display search results as you are typing. (google like)

https://jira.kaliop.net/browse/KZL-301

  - Step 1 : Create a index and collection and fill it with fake data (or use [this script](https://gist.github.com/Aschen/67ee985b2445273edea6b172414e9bfe) )
  - Step 2 : Go to the collection and type something in the simple filter input
  - Step 3 : Switch to basic filter and type something


Aschen:
I agree. I add the realtime search on quick filter because it was specified in the story but I'm not quite sure we have a strong usecase.

Aschen:
I have removed the realtime search result for Quickfilter

PR #431, opened 2018-08-20
Security refactor

xbill82:
## What does this PR do ?
~depends on #430~

MAKE THE ADMINCONSOLE GREAT AGAIN

This PR refactors the security lists (Users, Profiles and Roles) harmonizing the propagation of the filters with the Data section. 

### How should this be manually tested?

Just take a look around and see if anything is broken

### Boyscout

This PR excludes the `vuex-router-sync` package, that we wanted to discard.

PR #435, opened 2018-08-22
KZL-300 Buttons to switch between list views

xbill82:
~Depends on #431~

## What does this PR do ?
This PR enriches the Documents page (Data :arrow_right: `<index>` :arrow_right: `<collection>`)  with three buttons next to the QuickSearch filter. These buttons will allow to show the documents contained in the collection as two new views (Boxes and Map).

Open the contents of a collection and click on the three buttons next to the QuickSearch filter. 

selection_115

You should see two new views in the place of the document list.

selection_116

### Other changes
The PR introduces a new refactoring to the Documents Page.
The refactoring was necessary because the CommonList and CrudlDocument component introduced useless abstraction that reduced flexibility in the development of new features. These components are no more used on this page: instead, their contents have been


jenow:
For me the 3 buttons to switch the view are too big.. We need other opinions though

PR #436, opened 2018-08-23
Shrinked the main wrapper a little bit and centered it

xbill82:
## What does this PR do ?
This PR introduces a layout proposal for the content pane of the Admin Console. It's just a proposal and I'd like to ask the community for feedback.

### Motivation
Working on the Admin Console on large screens is annoying to me due to the item lists taking the whole screen width. I found that list items don't need all that space, since content is mostly vertically laid out and takes one fifth of the screen width at most (when the item content is expanded).
Plus, this layout makes it painful to reach the action buttons for items (the little pencil and the "Delete" menu item).
And, last but not least, it simply looked ugly to my eyes. Not that I find this new layout particularly pretty, but it seems to me it's on the right track.

Open the Netlify-deployed instance or build your own locally and connect it to a running Kuzzle Server. Just browse the documents of a collection or the Security section and see how it looks. You can also play with the window width and see how it behaves.

#### Before
kuzzle admin console - mozilla firefox_117

#### After
kuzzle admin console - mozilla firefox_118

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2371,7 +2371,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


xbill82:
LOL, this went too fast.

PR #439, opened 2018-08-23
[KZL-181] As an end user, i'd like to use Kuzzle Admin Console on my desktop

alexandrebouthinon:
:warning: Don't merge this PR yourself :warning: 

## What does this PR do ?
Add a simple electron packaging stage to Travis and push ZIP to dl.kuzzle.io

Check https://dl.kuzzle.io and download the zipfile matching your platform.


alexandrebouthinon:
WARNING : Do not merge this PR ! 
I use travis to publish the DoD on S3. So I need to update the travis.yml when all the reviewers accept this PR.

alexandrebouthinon:
@Aschen: It's Ok for the zoom level and version number in package name. But version number in window title must be perform in VueJS code not in Electron.

.electron/main.js
```diff
@@ -0,0 +1,57 @@
...
+let mainWindow;
+
+function createWindow () {
+  // Create the browser window.
+  mainWindow = new BrowserWindow({
+    width: 1280,
+    height: 720,
+    devTools: false,
+    webPreferences: {
+      zoomFactor: 0.8
```
xbill82: I think this is the DPI @benoitvidis is talking about: why not `1`?

alexandrebouthinon: Because, we need to scroll down to fill form pages. But I think have found a solution. I'll put the PR on WIP.

xbill82: it looks more like a CSS issue than something we should tweak on Electron, no?

alexandrebouthinon: Yes, you're right :disappointed:

PR #442, opened 2018-08-27
Document blocks list view

xbill82:
~depends on #441~

## What does this PR do ?
This PR introduces the document boxes view in the Document List. 

* Open the content of a non-empty collection in the Data section
* Click on the "Blocks" icon on the right of the quick search filter
* You should see the documents in form of blocks


src/components/Data/Documents/DocumentBoxItem.vue
```diff
@@ -0,0 +1,115 @@
...
+            <a
+              v-bind:class="{'disabled': !canDelete}"
+              @click="deleteDocument"
+              >
+              Delete
+            </a>
+          </li>
+        </dropdown>
+      </div>
+      <p v-for="(value, key) in documentContent" :key="value"><strong>{{key}}</strong>: {{value}}</p>
```
Aschen: The `key` used is the document content, maybe you can use the document id instead ? It's cause an error that prevent the component to render: `Avoid using non-primitive value as key, use string/number value instead.`

src/components/Data/Documents/Page.vue
```diff
@@ -99,11 +99,32 @@
             </div>
           </div>
 
-          <div class="DocumentList-boxes col s12" v-show="currentFilter.listViewType === 'boxes'">
-            <i class="fa fa-th fa-5x"></i>
-            <h2>Boxes List view</h2>
-            <p>This feature is not yet implemented.</p>
-            <p>Hold on, we'll ship it soon!</p>
+          <div class="col s12" v-show="listViewType === 'boxes'">
```
Aschen: `listViewType` does not exists any more, use `currentFilter.listViewType`

src/components/Data/Documents/DocumentBoxItem.vue
```diff
@@ -0,0 +1,115 @@
...
+      return canDeleteDocument(this.index, this.collection)
+    },
+    checkboxId() {
+      return `checkbox-${this.document.id}`
+    }
+  }
+}
+</script>
+
+<style type="scss" rel="stylesheet/scss" scoped>
```
Aschen: The document content leaves the box when it's too large, maybe you should add an `overflow: scroll`

image

src/components/Data/Documents/DocumentBoxItem.vue
Aschen: And why are you not using the json formatter directive ?
You display does well for simple documents but it feel weird with nested documents.

src/components/Data/Documents/DocumentBoxItem.vue
```diff
@@ -0,0 +1,115 @@
...
+              v-bind:class="{'disabled': !canEdit}"
+              @click="editDocument"
+              >
+              Edit
+            </a>
+          </li>
+          <li class="divider"></li>
+          <li>
+            <a
+              v-bind:class="{'disabled': !canDelete}"
```
Aschen: If the user is not allowed to delete the document, he will still see the action button but there will be no effect on click. Maybe it's better to disable it if the user is not allowed to do this.

Despite this design cause an inconsistency between the list and map view and the block view because we are using a button for editing and a dropdown with the delete action for list and map and just a dropdown for block.
[map view action button](https://github.com/kuzzleio/kuzzle-admin-console/pull/444/files#diff-919df2d4cb6f9e0a3a455aace5330b56R18)

src/components/Data/Documents/DocumentBoxItem.vue
```diff
@@ -0,0 +1,129 @@
...
+      return canDeleteDocument(this.index, this.collection)
+    },
+    checkboxId() {
+      return `checkbox-${this.document.id}`
+    }
+  }
+}
+</script>
+
+<style type="scss" rel="stylesheet/scss" scoped>
```
Njuelle: as we are in the process of refactoring also the CSS in components, I think we can use nested rules because it's easier to read

PR #444, opened 2018-08-27
KZL 305 - Visualize documents on a map

Aschen:
## What does this PR do ?

This PR add a ViewMap component to display documents on a map.  
The component use the documents `geo_point` to display markers on the map.  
The user can choose which `geo_point` to use with a dropdown.  
When you click on a marker, you can see the document content

  - Step 1 : Execute [this script](https://gist.github.com/Aschen/c7e7509aaa598dbe7d5b7efd4f3d4f68) to have some geolocated data
  - Step 2 : Go to the collection view and select map view in the top right corner
  - Step 3 : Select some marker, you should see information about the document
  - Step 4 : Change the selected geopoint to `emplacement` instead of `location`

### Boyscout

  - fix bug in the throttle that cause inconcistency between the quick filter content and the url

image


dev/templates/Component.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
xbill82: Lacks a comment here to explain what it is

src/components/Common/Filters/QuickFilter.vue
```diff
@@ -64,7 +64,7 @@ export default {
       this.throttleSearch = true
       setTimeout(() => {
         this.throttleSearch = false
-      }, 200)
+      }, 50)
```
xbill82: boyscouting?

src/components/Data/Documents/ListActions.vue
```diff
@@ -12,14 +12,12 @@
 
       <button class="btn btn-small waves-effect waves-light margin-right-5 primary"
         @click.prevent="$emit('create')"
-        :class="!displayCreate ? 'disabled' : ''"
-        :disabled="!displayCreate"
-        :title="displayCreate ? '' : 'You are not allowed to create a document in this collection'">
+      >
```
xbill82: I don't understand this.

src/components/Data/Documents/ListActions.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="ListActions row actions">
     <div class="col s8">
-      <button
+      <button v-if="viewType !== 'map'"
```
xbill82: In order to keep this component more agnostic and improve encapsulation, it's better to use boolean props that explicitly tell which buttons to display (e.g. `showToggleBtn`, `showBulkDeleteBtn`, etc...), then leave to the parent the logic that sets them to `true` or `false`.

src/components/Data/Documents/ListActions.vue
```diff
@@ -28,16 +26,46 @@
         Delete
       </button>
     </div>
+    <div v-if="viewType === 'map'" class="col s2">
```
xbill82: Wouldn't it be better to encapsulate this in a single `<div>` with just one `v-if`?

src/components/Data/Documents/Page.vue
```diff
@@ -238,6 +267,29 @@ export default {
     }
   },
   methods: {
+    getCoordinates(document) {
```
xbill82: Please, regroup the "geomap" methods with a big comment like the others. It's not very important but helps a little bit to keep this huge component tidy.

src/components/Data/Documents/Page.vue
```diff
@@ -412,12 +464,38 @@ export default {
...
+
+          attributes = attributes.concat(
+            this.listGeopoints(properties, path.concat(attributeName))
+          )
+        }
+      }
+
+      return attributes
+    },
+    loadCollectionInfo() {
```
xbill82: Since this method basically calls `getMappingDocument`, it essentially loads the ES mapping, right? I'm afraid the name `loadCollectionInfo` is a bit misleading...

xbill82: Also, didn't we already have a method that loads the mapping of the collection in this same page? If I'm correct, it was used to implement the basic-search field suggestions.

src/components/Data/Documents/ViewMap.vue
```diff
@@ -0,0 +1,209 @@
...
+        shadowUrl: 'static/images/marker-shadow.png',
+        iconSize: [25, 41],
+        iconAnchor: [12, 41],
+        popupAnchor: [1, -34],
+        shadowSize: [41, 41]
+      })
+    }
+  },
+  computed: {
+    latLngs() {
```
xbill82: I'd keep the `coordinates` name on all the methods (`latLngs` and `getCoordinates` use two names for the same concept)

src/main.js
```diff
@@ -1,6 +1,7 @@
 import Vue from 'vue'
-// import router from './services/router'
 import VueRouter from 'vue-router'
+// import Vue2Leaflet from 'vue2-leaflet'
```
xbill82: Can these comments be deleted?

src/main.js
```diff
@@ -9,6 +10,7 @@ import * as types from './vuex/modules/common/kuzzle/mutation-types'
 initStoreWithKuzzle(store)
 
 Vue.use(VueRouter)
+// Vue.use(Vue2Leaflet)
```
xbill82: Same as above

src/components/Common/Filters/QuickFilter.vue
Aschen: Yes, they were inconsistency between the form and the url with a big timeout

src/components/Data/Documents/ListActions.vue
Aschen: I simply copy/paste the existent code. The create button can be disabled when the user does not have the correct rights on the collection.

Aschen: This component was rendered here: https://github.com/kuzzleio/kuzzle-admin-console/pull/444/files/857e6bccb21e13d08dbe2a34ea344c408d5058a6#diff-0805784b9e3841d9b40f8950321b4e38L63

Aschen: It seems to be an error, I corrected that

src/components/Data/Documents/Page.vue
Aschen: It is the same method. Before the call to `getMappingDocument` was in the `mounted` method.  
I renamed it to `loadMappingInfo` because it loads the mapping and then find the geopoint list.

PR #447, opened 2018-08-29
Run prettier on the whole project

Aschen:
image

Me and @xbill82 were bored to have enormous diff at each pull request because all files were not prettiered.


Aschen:
No, strange

Issue #437, opened 2018-08-23
404 error when clicking on a collection name in the sidebar search


lethak:
## Expected Behavior
The dedicated collection page should open instead of a 404 error.

## Current Behavior
404 error when clicking on a collection name in the sidebar search

![screenshot-console kuzzle io-2018 08 23-11-25-15](https://user-images.githubusercontent.com/1705300/44517366-c9414200-a6c7-11e8-8a82-e5e728f1616b.png)

generated url: http://console.kuzzle.io/#/data/global/%3Cstrong%20class=%22highlight%22%3Eapplications%3C%2Fstrong%3E

correct url:
http://console.kuzzle.io/#/data/global/applications

## Steps to Reproduce
1. Open the admin console
2. Search for an existing collection name in the left sidebar
3. click on the found collection name under the index name (called "applications" in the previous screenshot)

## Context (Environment)
http://console.kuzzle.io

xbill82:
Can't reproduce this. Whenever I open an index I can properly see the collections stored inside.

xbill82:
Ok, now I can reproduce.

PR #454, opened 2018-09-17
HOTFIXed collection filter

xbill82:
## What does this PR do ?
This PR fixes the issue #437 as long as two other small fixes.

See #437.

### Other changes

* Fixed the position of the dropdown menu in the DocumentBoxedItem
* More aggressively stripping garbage characters in the internal `:id` of Materialize Dropdown


xbill82:
@scottinet well, I think you're right. It doesn't really deserve a hotfix and indeed it should be merged with the other one.

PR #459, opened 2018-09-20
Fix search of nested geopoints

Aschen:
## What does this PR do ?

For the map view, we are looking for a `geo_point` property in the document. But the function does not support nested `geo_point`.

This PR add support for nested `geo_point` detection. It also consider a property that have two nested properties called `lat` and `lon` as a geo compatible property.

In this example, both `home` and `work` properties are geo compatible properties:
```js
{
   home: { type: 'geo_point' },
   work: {
     properties: {
       lat: { type: 'long' },
       lon: { type: 'long' }
     }
}
```

  - Step 1 : Use this [script](https://gist.github.com/Aschen/7825c8160cb9b7aaf7e1cc0e4ffa1197) to create some documents
  - Step 2 : Go to the map view and try to switch between `location` and `nested.emplacement`

### Other changes

 - Fix geo point selector
 - Update docker-compose

### DIFF
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -67,7 +67,7 @@
               v-if="documents.length"
               :all-checked="allChecked"
               :display-bulk-delete="hasSelectedDocuments"
-              :mappingGeopoints="mappingGeopoints"
+              :geopointList="mappingGeopoints"
               :viewType="currentFilter.listViewType"
               :displayCreate="canCreateDocument(this.index, this.collection)"
               :displayGeopointSelect="currentFilter.listViewType === 'map'"
@@ -329,7 +329,7 @@ export default {
           if (properties.lat && properties.lon) {
             attributes = attributes.concat(path.concat(attributeName).join('.'))
           }
-          
+
           attributes = attributes.concat(
             this.listMappingGeopoints(properties, path.concat(attributeName))
           )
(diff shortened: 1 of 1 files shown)


src/components/Data/Documents/Page.vue
```diff
@@ -324,8 +324,16 @@ export default {
...
 
-      for (const [attributeName, { type }] of Object.entries(mapping)) {
-        if (type === 'geo_point') {
+      for (const [attributeName, { type, properties }] of Object.entries(mapping)) {
+        if (properties) {
+          if (properties.lat && properties.lon) {
+            attributes = attributes.concat(path.concat(attributeName).join('.'))
+          }
+
+          attributes = attributes.concat(
```
scottinet: Why create a new array from scratch every time we want to add a value to the local `attributes` array, instead of just pushing the value in the existing array?

`attributes.push(this.listMappingGeopoints(properties, path.concat(attributeName)));`

Aschen: Because `listMappingGeopoints` return an array and I will end up with an array of array

scottinet: You can always spread it: `attributes.push(...this.listMappingGeopoints(properties, path.concat(attributeName)));`

Aschen:
WIP because I just found this: https://docs.kuzzle.io/kuzzle-dsl/essential/geopoints

PR #473, opened 2018-10-25
KZL-694 Refresh search when hitting "Search"

### LINKED ISSUE #467: Playing the same advanced query doesn't update the result set
I'm in the admin console I'm running the following query : 

```json 
{
  "query": {
    "prefix": {
      "image.uri": "https://res.cloudinary.com/dkiv4hwn3/image/upload/v"
    }
  }
}
```

## Expected Behavior

I click on `Search` and I get a result set indicating X documents.
I run a script in another place that is supposed to the reduce this set matching this query.
I click again,  same amount of documents matching displayed at the bottom of the page.
I reload the page, this replay the query and the correct amount is displayed.

## Possible Solution
The query is probably not sent.

## Context (Environment)

I'm using **admin console 2.2.2**

Best regards

xbill82:
## What does this PR do ?
fixes #467 
Before, the DocumentList page watched the `currentFilter` object for changes to trigger the `fetchDocuments` function. I changed this behavior, now the `fetchDocuments` function is triggered every time the "search" button is clicked.

1. Launch local End to End tests with `npm run e2e-local`
2. Start the `search.spec.js` spec file
3. See what happens in the `refreshes search when the Search button is hit twice` test case
4. Reproduce it in your favorite browser.

### Other changes
The test framework was a little bit broken. Now it's meant to be more stable.


src/components/Data/Documents/Page.vue
```diff
@@ -563,9 +565,9 @@ export default {
         )
       }
     },
-    currentFilter() {
-      this.fetchDocuments()
-    },
+    // currentFilter() {
```
etrousset: we should not leave commented code

PR #488, opened 2018-11-29
KZL-581 Translate Basic search queries to Raw

xbill82:
## What does this PR do ?
This PR introduces the feature to see a query made via the Basic filter in the Raw filter tab, so that the user can see the actual Elasticsearch specification for the query. This is useful to learn the Elasticsearch DSL but also to have a simple way to generate queries that the user can enrich manually later.

1. Log into a Kuzzle with some data
2. Create a Basic filter
3. Click the "Translate to Raw" button
4. You should see the Raw filter editor filled with the query.

### Boyscout

The new version of Prettier installed in my VSCode just formatted the `template` part of the files that I have touched.

### DIFF
--- a/src/components/Common/Filters/BasicFilter.vue
+++ b/src/components/Common/Filters/BasicFilter.vue
@@ -103,9 +103,7 @@
           class="BasicFilter-translateBtn btn waves-effect waves-light success right"
           v-bind:disabled="!isFilterValid"
           @click.prevent="translateToRaw"
-        >
-          <i class="fa fa-angle-double-right"></i> Translate to RAW
-        </button>
+        >Translate to RAW</button>
       </div>
     </div>
   </form>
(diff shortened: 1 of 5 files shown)


Aschen:
Good work !! Very nice feature for Elastic newcomers :tada: 

Just a little byg, after a click to "Translate to Raw" I can't go back to the basic mode.
peek 2018-11-30 15-50

And why don't just translate the query in realtime in the "RAW JSON Mode" instead of having to click on "Translate to raw" .

xbill82:
@Aschen Good points.
The feature of translating in real-time was implemented two years ago and it resulted in a messy UX in the case the user had pre-filled a query in the Raw tab, then filled some fields in the Basic search... It changed the Raw search without any visual feedback. We decided to discard it after trying a it little bit.
The problem of not being able to get back to the Basic filter is a bug related to the Tabs MaterializeCSS component, which is completely messy and very costly to fix. The ideal solution would be to wipe Materialize away one time for all, but I'm afraid I should put this PR in WIP state and just fix that crap.

xbill82:
@benoitvidis I agree with your point, I think this should be optimized. Although, this is not in the scope of this feature, I see it more in the mere generation of the query (refer to the `formatBasicSearch()` in `filterManager.js`) and therefore should be made in a separate PR.

Issue #493, opened 2019-02-07
User-Rights :  Profil restrictions to specific index/collection doesnt apply properly to treeview


Clembl-git:
After applying a restriction ( "restrictedTo": [ { "index": "myIndex"}] ) to a profil, the associated user cant list any items on the indexes treeview of the admin console.  The error message "you are not allowed to list indexes" appear, **even if the access through the url still works (console.kuzzle.io/#/data/myIndex).** 

## Expected Behavior
The "list" action of the "index" controller should take account of the potentials restrictions defined in the corresponding profile.

## Current Behavior
If i restrict a user to a specific index, all system indexes (watcher-history, triggered_watches, monitoring-alert..) are still visible. If i define a specific role (lets say restrictedIndex) define like this: 
{
  "controllers": {
    "index": {
      "actions": {
        "list": true
      }
    }
}
and then if i restrict my profile like bellow :  {
"roleId": "restrictedIndex",
 "restrictedTo": [{"index": "inventory"}] 
}
All indexes disappear from the treeview. Same behavior with an admin profile and role.

## Possible Solution
The "canSearchIndex()" method in userAuthorization.js should take the list of restrictions into account

PR #494, opened 2019-02-07
Fix 493 list only restricted index

Aschen:
## What does this PR do ?

Allows to list indexes even if the current user is restricted to some.

Fix: https://github.com/kuzzleio/kuzzle-admin-console/issues/493

  - Step 1 : Create an user with a profile restricted to an index
  - Step 2 : You should see only the restricted index

### DIFF
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -72,7 +72,7 @@ export default {
             collection: this.$route.params.collection
           })
         })
-        .catch(err => console.error(err))
+        .catch(() => null)
     }
   },
   watch: {
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -72,7 +72,7 @@ export default {
             collection: this.$route.params.collection
           })
         })
-        .catch(() => null)
+        .catch(error => console.error(error))
     }
   },
   watch: {
--- a/src/services/userAuthorization.js
+++ b/src/services/userAuthorization.js
@@ -22,7 +22,10 @@ function isActionAllowed(
 // Index CRUDL
 
 export const canSearchIndex = () => {
-  const indexListRight = store.state.user.rights.filter(rights => rights.action === 'list' && rights.controller === 'index')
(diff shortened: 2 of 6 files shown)


src/services/userAuthorization.js
```diff
@@ -22,7 +22,12 @@ function isActionAllowed(
 // Index CRUDL
 
 export const canSearchIndex = () => {
-  return isActionAllowed(store.state.user, 'index', 'list')
+  const indexListRight = store.state.user.rights.filter(rights =>
```
scottinet: From the look of it, it seems you're replacing a faulty `isActionAllowed` function with some custom code.

Why isn't `isActionAllowed` returning `allowed` on `index:list`, as it should be?

Aschen: The function `isActionAllowed` does not have sense for `index:list` and `collection:list` because these actions does not take in account the index and collection restricted in the user profile.

src/vuex/modules/index/actions.js
```diff
@@ -61,16 +61,16 @@ export default {
               resolveOne(indexesAndCollections)
             })
           })
+            .then(res => {
+              commit(types.RECEIVE_INDEXES_COLLECTIONS, res || {})
+            })
+            .catch(() => null) // Silent on purpose when listCollection is forbidden
```
xbill82: Silent in case this action is forbidden, but what if there's a real error?

src/components/Data/Layout.vue
```diff
@@ -72,7 +72,7 @@ export default {
             collection: this.$route.params.collection
           })
         })
-        .catch(err => console.error(err))
+        .catch(() => null)
```
xbill82: Is it OK to hide errors to the user?

Aschen: Error thrown here are Unauthorized errors for list collections on restricted indexes

xbill82: How can you be 100% sure the _only_ errors thrown here are related to rights?

Aschen: Fixed

PR #502, opened 2019-03-07
Add ability to define mapping policy for new fields

Aschen:
## What does this PR do ?

Allows the definition of the `dynamic` property for the mappings.  
Now users have to specify the root `property` field for mappings:
```
{
  "dynamic": "true",
  "properties": {
    "age": { "type": "integer" },
    "name": { "type": "text" }
  }
}
```

This PR is related to https://github.com/kuzzleio/kuzzle/pull/1257 but the admin console will work with older kuzzle version too.

  - Step 0 : Run Kuzzle with https://github.com/kuzzleio/kuzzle/pull/1257
  - Step 1 : Create a collection and set the `dynamic` property to `strict` at the same level as `properties`
  - Step 2 : Try to create a document with a field that is not in the mapping

### Other changes

  - change error text to white


src/components/Data/Collections/CreateOrUpdate.vue
```diff
@@ -14,7 +14,7 @@
       :steps="['Mapping', 'Form']"
       @changed-step="setEditionStep"
       class="card-panel card-header">
-    </stepper>
+    </stepper>²
```
Njuelle: I didn't know we could use ² to multiply div in html :D
```suggestion
    </stepper>
```

src/components/Data/Documents/Common/CreateOrUpdate.vue
```diff
@@ -86,7 +86,7 @@
           <div class="col s7 m8 l9" v-if="error">
             <div class="card error red-color">
               <i class="fa fa-times dismiss-error" @click="dismissError()"></i>
-              <p v-html="error"></p>
+              <p class="white-text" v-html="error"></p>
```
Njuelle: You add a `white-text` class here, and the css rule apply to `.error-text`. is it normal? 
(the same for each error card you modified)

src/components/Materialize/Toaster.vue
```diff
@@ -8,6 +8,7 @@ export default {
   watch: {
     '$store.state.toaster.toast'() {
       /* eslint no-undef: 0 */
+      console.log(this.$store.state.toaster.toast)
```
Njuelle: ;)
```suggestion
```

src/components/Data/Documents/Common/CreateOrUpdate.vue
```diff
@@ -143,6 +143,9 @@
     padding: 8px 12px;
     margin: 0;
   }
+  .error-text {
```
Njuelle: see previous

PR #504, opened 2019-03-08
Port materialize css to 1.0.0

jenow:
## What does this PR do ?

Port materialize CSS to version 1.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "kuzzle-sdk": "^5.0.11",
     "lodash": "^4.17.10",
     "lolex": "^1.5.2",
-    "materialize-css": "^0.100.2",
+    "materialize-css": "^1.0.0",
     "mocha": "^5.2.0",
     "moment": "^2.22.1",
     "node-sass": "^4.9.2",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kuzzle-admin-console",
-  "version": "2.2.4",
+  "version": "2.3.4",
```
scottinet: why do you change the version number manually to release on the development branch?

scottinet: Also, if you want the minor version to be updated, then you should use the `changelog:new-features": this triggers a minor version update from our release tools.

jenow: I don't want to update the last number as we change it for every hotfix, I want to change the second one. It should be 2.3.0 btw -.-

scottinet: You don't get it: if you manually update the version and release the admin console with our tools, you'll get a "2.3.1" version tag (unless you use `--tag` to force the future version number to 2.3.0).
If you leave the version alone and simply add the `changelog:new-features` label, then the release tools will automatically increment the minor version number.

jenow: Right, I'm sorry I did some hotfix before and I was used to change the version, forgetting I was making a PR on the development branch -_-

PR #505, opened 2019-03-11
[HOTFIX] Add a warning when a filter is set when looking at documents and users

### LINKED ISSUE #333: Show some filters exist even if options are closed
selection_036

If we put some conditions in the query options form, then we close the panel, those filters are not taken into account, but if we re-open the panel, they are present. So filters are actually there. Yet we don't see the pill saying there's some complex filter under the "More query options".

Moreover, the upper, light-blue "SEARCH" button does not work when complex filters are open.

jenow:
## What does this PR do ?

When a filter is set when looking at documents it is not abvious and it happens often for user to think there are missing documents but really they are filtered so I added a warning when a filter is set.

fix #333 

## Other

Removed the patch rule from codecov as it is useless and run in a forever loop anyway.

## Note

I didn't change the version as I am going to do a batch of hotfix before patching this version

### DIFF
--- a/src/components/Common/Filters/Filters.vue
+++ b/src/components/Common/Filters/Filters.vue
@@ -55,7 +55,7 @@
     </div>
 
     <div class="card-panel orange lighten-3" v-show="currentFilter.active">
-      <span>Warning: There is actually an active filter, not every documents are shown.</span>
+      <span>Warning: a filter has been set, some documents might be hidden.</span>
     </div>
 
   </div>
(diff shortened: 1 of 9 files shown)


src/components/Common/Filters/Filters.vue
```diff
@@ -53,7 +53,13 @@
         </div>
       </tabs>
     </div>
+
+    <div class="card-panel orange lighten-3" v-show="currentFilter.active">
+      <span>Warning: There is actually an active filter, not every documents are shown.</span>
```
scottinet: The way I understand it, you meant `currently` and not `actually`, which is a false friend.

Here is a change proposal, I let you decide how you want to deal with this.

```suggestion
      <span>Warning: a filter has been set, some documents might be hidden.</span>
```

PR #515, opened 2019-03-22
Range / Lower / Greater filter

### LINKED ISSUE #466: Allow to use gte / lte and other like filter in search
# Add capability to filter with greater / lower than.

In kuzzle admin console, it could be nice to have « math » operators for the search panel: 

<a href="https://ibb.co/csCb69">Screenshot_2018_10_10_at_16_23_54</a>

Currently there is only `match` and `equals`.

Best regards

jenow:
## What does this PR do ?

Add the `range` filter on basic search component.

a

If you set **2** values it will return all documents wich have this value in **between** (not equal).
If you only set the **first** value then it will behave as a **greater** filter (not equal).
If you only set the **second** value thenit will behave as a **lower** filter (not equal).

## Others

Fix a minor error on the Watch component.

fix #466 

## Preview

https://deploy-preview-515--kuzzle-backoffice-last.netlify.com/

### DIFF
--- a/src/services/filterManager.js
+++ b/src/services/filterManager.js
@@ -318,7 +318,7 @@ export const formatFromBasicSearch = (groups = [[]]) => {
           }
         })
       } else if (filter.operator === 'range') {
-        let range = {range: {}}
+        const range = {range: {}}
         if (filter.gt_value && filter.lt_value) {
           range.range = {
             [filter.attribute]: {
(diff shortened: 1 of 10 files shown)


src/services/filterManager.js
```diff
@@ -316,6 +317,29 @@ export const formatFromBasicSearch = (groups = [[]]) => {
             }
           }
         })
+      } else if (filter.operator === 'range') {
+        let range = {range: {}}
```
scottinet: (nitpicking) 

```suggestion
        const range = {range: {}}
```

PR #516, opened 2019-03-22
Show basic filter in the raw view if active

### LINKED ISSUE #410: [Improvement] Display query from graphic interface to json interface
## Expected Behavior
When i build a search query with the graphic interface, having the query build in JSON at same time (after submit, not necessary in real-time) would be a good add-on. For a beginner in Elastic Search (like me ^^), sometimes it's difficult to build a correct query with all "should", "must", "bool" etc...

## Possible Solution
I think it's a good support to on left-side the query helper, and on the right-side the result of the query builded.

### LINKED ISSUE #461: Add Capability to switch to raw JSON without losing the query
# Add Capability to switch to raw JSON without losing the query

Currently when you switch to RAWJson you're losing the query you're building. Could be nice to switch without losing.

Feature description

Regards

jenow:
## What does this PR do ?

If any basic filter is set then we show it's JSON version in the raw filter view.
Also display an info saying why is there a content in the raw filter json form and explain that if changed, it won't change the basic filter view.

a

fix #461 
fix #410 

## Preview

https://deploy-preview-516--kuzzle-backoffice-last.netlify.com/


xbill82:
Duplicate of #488 (it's ok, this implementation is simpler and avoids the weird behaviour introduced in the UI when switching tabs programmatically)

xbill82:
Peek 2019-03-25 12-40

jenow:
> Duplicate of #488 (it's ok, this implementation is simpler and avoids the weird behaviour introduced in the UI when switching tabs programmatically)

Man there was 2 github issue and 1 Jira issue about this and none of them were assigned to you -.-

jenow:
> Integration tests are missing

I know we didn't discuss about this yet but if we decide to develop the v3 of this admin console I'd prefer to start writing e2e tests on the v3. If you want we can wait until we have the roadmap meeting this week to decide.

xbill82:
Yeah, my PR was done before the tickets and forgot in a corner. But I actually prefer your implem, so it's OK for me.
Concerning the tests, I'd like that we write all the test as long as we develop. They will help us developing the V3.

PR #517, opened 2019-03-25
[KZL-1004] Securities helpers

jenow:
## What does this PR do ?

Add an info alert when creating/editing a role/profile/user with a link to security documentation of Kuzzle.
Also add default exemple json when creating a role or profile.

## Other

- Don't show the `Document-creation form is not enabled for this collection` alert when creating / updating security related things.
- Remove unused CreateOrUpdate security component

a


src/components/Security/Common/Notice.vue
```diff
@@ -0,0 +1,17 @@
+<template functional>
+  <div class="SecurityNotice card-panel blue lighten-4">
+    For more information about securities policies please visit <a class="blue-text lighten-2" href="https://docs.kuzzle.io/guide/1/essentials/security/" target="_blank">https://docs.kuzzle.io/guide/1/essentials/security <i class="fa fa-external-link-alt"></i></a>
```
xbill82:
```suggestion
    For more information about security policies please visit <a class="blue-text lighten-2" href="https://docs.kuzzle.io/guide/1/essentials/security/" target="_blank">https://docs.kuzzle.io/guide/1/essentials/security <i class="fa fa-external-link-alt"></i></a>
```

src/components/Security/Profiles/Create.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <div>
     <Headline>Profile - Create</Headline>
+    <Notice></Notice>
```
xbill82: `Notice` is French, isn't it? How do we call this in English? :thinking:

xbill82: Dunno... we can leave this as is

PR #518, opened 2019-03-26
KZL-719-human-readable-datefields

berthieresteban:
## What does this PR do ?
In a document details, change the values of the date type fields (if timestamp) to add human readable date.

  - Step 1 : Run kuzzle
  - Step 2 : Create a document with date field 
  - Step 3 : Look at meta createdAt and the date field in admin console

### Other changes
/

### Boyscout
/

### DIFF
--- a/src/components/Data/Documents/DocumentListItem.vue
+++ b/src/components/Data/Documents/DocumentListItem.vue
@@ -71,6 +71,10 @@ export default {
       collapsed: true
     }
   },
+  mounted() {
+    const date = new Date(this.document.meta.createdAt)
+    this.document.meta.createdAt += ' (' + date.toUTCString() + ')'
+  },
   methods: {
     toggleCollapse() {
       this.collapsed = !this.collapsed
--- a/src/components/Data/Documents/DocumentListItem.vue
+++ b/src/components/Data/Documents/DocumentListItem.vue
@@ -73,7 +73,7 @@ export default {
   },
   mounted() {
     const date = new Date(this.document.meta.createdAt)
-    this.document.meta.createdAt += ' (' + date.toUTCString() + ')'
+    this.document.meta.createdAt += ` (${date.toUTCString()})`
   },
   methods: {
     toggleCollapse() {
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -588,7 +588,7 @@ export default {
     addHumanReadableDateFields() {
       const keys = []
       const findDateFields = (mapping, previousKey) => {
-        for (let key in mapping) {
+        for (const key of Object.keys(mapping)) {
           if (typeof mapping[key] === 'object') {
             findDateFields(mapping[key], key)
           } else if (key === 'type' && mapping[key] === 'date') {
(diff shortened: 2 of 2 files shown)


src/components/Data/Documents/DocumentListItem.vue
```diff
@@ -71,6 +71,10 @@ export default {
       collapsed: true
     }
   },
+  mounted() {
+    const date = new Date(this.document.meta.createdAt)
+    this.document.meta.createdAt += ' (' + date.toUTCString() + ')'
```
Aschen: Use string interpolation rather than concatenation

src/components/Data/Documents/Page.vue
```diff
@@ -584,6 +584,32 @@ export default {
         otherQueryParams
       )
       this.$router.push({ query: mergedQuery })
+    },
+    addHumanReadableDateFields() {
+      const keys = []
+      const findDateFields = (mapping, previousKey) => {
+        for (let key in mapping) {
```
Aschen: Use a `for (const .. of ...)` loop instead (https://alligator.io/js/for-of-for-in-loops/)

src/components/Data/Documents/Page.vue
```diff
@@ -584,6 +584,32 @@ export default {
...
+          if (typeof mapping[key] === 'object') {
+            findDateFields(mapping[key], key)
+          } else if (key === 'type' && mapping[key] === 'date') {
+            keys.push(previousKey)
+          }
+        }
+      }
+      const changeField = (document, keys) => {
+        for (let field in document) {
```
Aschen: Same remark about the `for` here

src/components/Data/Documents/Page.vue
```diff
@@ -584,6 +584,32 @@ export default {
...
+          if (keys.includes(field) && Number.isInteger(document[field])) {
+            const date = new Date(document[field])
+            document[field] += ' (' + date.toUTCString() + ')'
```
Aschen: Same remark about string interpolation

src/components/Data/Documents/Page.vue
berthieresteban: I need the key, if I use a for of i can't access to the key right ?
The site say :
Use for…in to iterate over the properties of an object (the object keys):
Could you explain me ?

src/components/Data/Documents/DocumentListItem.vue
berthieresteban: :+1:

src/components/Data/Documents/Page.vue
berthieresteban: :+1:

src/components/Data/Documents/Page.vue
Aschen: Oh yes sorry, usually we do this `for (const key of Object.keys(mapping))` but I agree that `for (let ... in ...)` work pretty the same here.  
The only differences are kind of nitpicking:
 - `const` instead of `let` ensure your can't reassign variable
 - more explicit iteration over keys (IMHO)
 - consistency between all for loop

berthieresteban: Thanks, i'll do the changes :+1:

PR #523, opened 2019-04-01
Save sort filter in raw json mode

jenow:
## What does this PR do ?

Save "sort" filter from a raw query in json mode.

- open the admin console to see the content of a collection
- open "query options"
- switch to "raw json mode"
- add a query and sort filter like:
```json
{
  "query": {},
  "sort": {"foo": "asc"}
}
```
- click on search
- go into another collection
- go back to the first collection under the "raw json mode" search options
- sort should still be there

a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,73 +5,73 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.51",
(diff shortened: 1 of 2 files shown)


jenow:
> Unstage the package-lock.json file

The package-lock.json is broken for some reason, the version of materialize-css is not hte right one so everything os broken.
I staged it on purpose in this commit https://github.com/kuzzleio/kuzzle-admin-console/pull/523/commits/2332ab793e1acbaaef803616585cdb30be8016a3

PR #527, opened 2019-04-25
Remove unauthorized actions

benoitvidis:
## What does this PR do ?

This PR removes unauthorized actions from the admin console, at least for read-only access.

1. Create a read-only user with the following role: 

```
{
  "controllers": {
    "auth": {
      "actions": {
        "getCurrentUser": true,
        "getMyCredentials": true,
        "getMyRights": true,
        "logout": true
      }
    },
    "collection": {
      "actions": {
        "getMapping": true,
        "list": true
      }
    },
    "document": {
      "actions": {
        "count": true,
        "get": true,
        "mGet": true,
        "scroll": true,
        "search": true
      }
    },
    "index": {
      "actions": {
        "list": true
      }
    }
  }
}
```
2. log in to the admin console
3. check ~no write action is available from~ all write actions are greyed out in the interface


scottinet:
I'm very bad at this, but I remember reading that hiding actions that cannot be performed is not a good UX practice. Instead, the article I read stated that it was better to always show all the possible actions, just greyed out for unavailable ones (or any convention showing clearly that something cannot be used right now).

That was a few years ago and, again, UX is not really my forte, so I don't know if that's relevant.

benoitvidis:
Seb is right => wip

src/components/Data/Collections/Boxed.vue
```diff
@@ -28,15 +28,15 @@
...
            :class="{unauthorized: !canCreateDocument(index, collection)}"
-		       :to="{name: 'DataCreateDocument', params: {index: index, collection: collection}}"
-           v-if="!isRealtime && canCreateDocument(index, collection)">
+		       :to="canCreateDocument(index, collection) ? {name: 'DataCreateDocument', params: {index: index, collection: collection}}: ''"
+           v-if="!isRealtime">
           Create a document
         </router-link>
         <router-link class="btn btn-small"
-           v-if="isRealtime && canManageRealtime(index, collection)"
+           v-if="isRealtime"
```
thomasarbona: You can use `v-else` directive here

benoitvidis: :+1:

PR #528, opened 2019-05-06
Show aggregations in search result if any

jenow:
## What does this PR do ?

In raw filter mode the user can set aggregations to the search but we didn't show them back in the search result.

- Go to raw json search in a collection.
- Enter a filter like this one:

```
{
  "query": {},
  "aggregations": {
    "my_aggs": {
      "terms": {
        "field": "foo"
      }
    }
  }
}
```

- Hit Search

In the result list you should see the `Content`, `Meta` and newly added `Aggregations` inside the documents.

## Preview

https://deploy-preview-528--kuzzle-backoffice-last.netlify.com/


src/services/kuzzleWrapper.js
```diff
@@ -121,12 +134,16 @@ export const performSearchDocuments = async (
   }
 
   const documents = result.hits.map(document => {
-    const object = {
+    let object = {
```
scottinet: should be const

test/e2e/cypress/integration/search.spec.js
```diff
@@ -11,6 +11,7 @@ describe('Search', function() {
       properties: {
         firstName: {
           type: 'text',
+          fielddata: true,
```
scottinet: shouldn't this be camel-cased?

test/e2e/cypress/integration/search.spec.js
```diff
@@ -364,4 +368,77 @@ describe('Search', function() {
       .and('contain', 'match_phrase_prefix')
       .and('contain', 'bar')
   })
+
+  it.only('should show aggregations in search result when aggregations are specified in the raw filter', function() {
```
scottinet: `it.only`: debugging leftover?

test/e2e/cypress/integration/search.spec.js
jenow: No it's an elasticsearch prop: https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html

src/services/kuzzleWrapper.js
jenow: I change this object in case there is aggregation:
```
    if (result.aggregations) {
      object.aggregations = new Aggregations(result.aggregations)
    }
```

scottinet: You mutate it, but this is still the same object.

REPL example:

```
$ node
> const foo = {bar: 'baz' }
undefined
> foo.qux = true
true
> foo
{ bar: 'baz', qux: true }
> 
```

jenow: javascript....

scottinet: For once, I disagree on that one. An object acts a bit like a hashmap: when you add a key, you mutate the container, but that doesn't change its address, so its fine.
If you want a constant object, you have to [freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) it (and all other objets/arrays in it obviously, for the same reason).

Same goes for arrays: you can have a const array and still push values in it.

PR #534, opened 2019-05-16
Column list view

jenow:
## What does this PR do ?

Add a new Column list view to view document and be able to select which column we want to see.
Those column will be saved in the local storage.

column


src/components/Data/Documents/DocumentListViewColumn.vue
```diff
@@ -0,0 +1,293 @@
...
+  directives: {
+    Title,
+    JsonFormatter
+  },
+  components: {
+    Dropdown,
+    Autocomplete
+  },
+  mounted () {
+    const columnsConfig = JSON.parse(localStorage.getItem('columnViewConfig') || '{}')
```
alexandrebouthinon: I'm not sure but I think if you are in "private navigation" on your browser, it's the `sessionStorage` that is filled instead of `localstorage`. It can be good to have a fallback to sessionStorage before the empty object :thinking:

src/components/Data/Documents/DocumentListViewColumn.vue
```diff
@@ -0,0 +1,293 @@
...
+      }
+
+      return {
+        isObject: false,
+        value: ret
+      }
+    },
+    saveToLocalStorage() {
+      if (this.index && this.collection) {
+        const config = JSON.parse(localStorage.getItem('columnViewConfig') || '{}')
```
alexandrebouthinon: See above

src/components/Data/Documents/DocumentListViewColumn.vue
```diff
@@ -0,0 +1,293 @@
...
+      }
+      return canDeleteDocument(this.index, this.collection)
+    },
+    checkboxId() {
+      return `checkbox-${this.document.id}`
+    }
+  },
+  watch: {
+    '$route' (to, from) {
+      const columnsConfig = JSON.parse(localStorage.getItem('columnViewConfig') || '{}')
```
alexandrebouthinon: See above

src/components/Data/Documents/DocumentListViewColumn.vue
jenow: It's on purpose, I want to have a brand new environment when I browse in private.

src/components/Data/Documents/DocumentListViewColumn.vue
```diff
@@ -0,0 +1,293 @@
...
+  methods: {
+    toggleJsonFormatter(id) {
+      if (this.$refs[id][0].style.visibility === 'hidden') {
+        this.$refs[id][0].style.visibility = 'visible'
+      } else {
+        this.$refs[id][0].style.visibility = 'hidden'
+      }
+    },
+    parseDocument(attr, doc) {
+      let ret = attr.includes('.') ? this.getNestedField(doc.content, attr) : doc.content[attr]
```
scottinet:
```suggestion
      const ret = attr.includes('.') ? this.getNestedField(doc.content, attr) : doc.content[attr]
```

src/components/Data/Documents/Page.vue
```diff
@@ -648,7 +682,7 @@ export default {
   }
 }
 .DocumentsPage-container {
-  max-width: $container-width;
+  // max-width: $container-width;
```
scottinet: to be deleted?

src/components/Data/Documents/DocumentListViewColumn.vue
```diff
@@ -0,0 +1,293 @@
...
+
+      if (typeof ret === 'object') {
```
scottinet: Can ret be null? Because if it can, then a `TypeError` will occur because the code will try to run `parseDocument` on null

src/components/Common/Autocomplete.vue
```diff
@@ -6,7 +6,7 @@
       :class="inputClass"
       :placeholder="placeholder"
       @input="onInput"
-      @change="(evt) => changeResult(evt.target.value)"
+      @change="evt => { if (this.notifyChange) { return changeResult(evt.target.value)}}"
```
thomasarbona: Not terrible to have so many logic in template part

src/components/Common/Autocomplete.vue
```diff
@@ -68,7 +73,7 @@ export default {
   },
   methods: {
     onInput() {
-      if (this.results.length > 0) {
+      if (this.results.length >= 0) {
```
thomasarbona: So, always true ?

PR #538, opened 2019-06-14
Time series view

jenow:
## What does this PR do ?

Add a time serie view where you can select a field of type `date` as the date reference and then add fields from you collection of the following type:

- short
- integer
- long
- double
- float
- half_float
- scaled_float
- binary
- byte

It will then display a graph with the different data.

To display data between 2 dates you can use the `range` filter of the `quick search`.

charts

## Preview

https://deploy-preview-538--kuzzle-backoffice-last.netlify.com/

# Boyscout

- fix a problem on the `discarded` event where sometimes the Vue app would not be yet initialized.
- Added the value 500 on the `result per page` option


jenow:
> Very nice :)
> 
> Some nipickings:
> 
> * The color pickers are a little big to me. I'd rather see them next to the field name
> * Once a first series is picked up, would it be possible to filter the fields list to only display compatible types?
> * Would it be possible by default to have a different color for each series?
> 
> A little more problematic for me:
> 
> * There is no scale / legend for the x-date axis
> * ~Shouldn't we have a date-range selector somewhere?~ <- sorry, I had not read the description about the use of the filters

- I will see what I can do.
- I already filter compatible types, they are listed in the description of this PR.
- I could set a random color for each series but we might see weird color being picked up :p
- I don't really understand what you mean, you can change the scale by scrolling with the mouse. What would you suggest?

benoitvidis:
About the missing scale: Here is what I get on netlify and on the dev branch on my system
image
I cannot see any scale for the x axis.

jenow:
@benoitvidis I think that's because you chose an attribute where the value in your documents is not defined. If you choose `timeOfReport` you should see it.

src/components/Data/Documents/Views/TimeSeries.vue
```diff
@@ -0,0 +1,270 @@
...
+import Autocomplete from '../../../Common/Autocomplete'
+import TimeSeriesItem from './TimeSeriesItem'
+import { GChart } from 'vue-google-charts'
+import _ from 'lodash'
+
+const ES_NUMBER_DATA_TYPE = [
+  'short',
+  'integer',
+  'long',
+  'doube',
```
scottinet:
```suggestion
  'double',
```

PR #542, opened 2019-06-17
[HOTFIX] Render search results with undefined meta

xbill82:
## What does this PR do ?
Fixes issue that broke document view when some of the documents had undefined `_meta` field.

Serve some documents with the `_meta` field set to `undefined` (you can use a pipe plugin in your Kuzzle backend) and see them in the documents page.

### Other changes
Show the error traces after showing the toast.


src/components/Data/Documents/Page.vue
```diff
@@ -486,6 +486,7 @@ export default {
         this.deleteModalIsLoading = false
       } catch (e) {
         this.$store.commit(SET_TOAST, { text: e.message })
+        console.error(e)
```
alexandrebouthinon: Unwanted `console.error` ?

src/components/Data/Documents/Page.vue
```diff
@@ -518,11 +519,12 @@ export default {
...
         )
         await this.fetchDocuments()
-      } catch (error) {
+      } catch (e) {
         this.$store.commit(SET_TOAST, {
           text:
-            'An error occurred while updating filters: <br />' + error.message
+            'An error occurred while updating filters: <br />' + e.message
         })
+        console.error(e)
```
alexandrebouthinon: Same ?

src/components/Data/Documents/Page.vue
```diff
@@ -560,6 +562,7 @@ export default {
           text:
             'An error occurred while performing search: <br />' + e.message
         })
+        console.error(e)
```
alexandrebouthinon: Same ?

src/components/Data/Documents/Page.vue
xbill82: Wanted.
We need to see the errors in order to debug them. It is therefore completely useless to display the message in the toast without having the trace in the console...

src/components/Data/Documents/Page.vue
xbill82: See above.

src/components/Data/Documents/Page.vue
xbill82: See above.

src/components/Data/Documents/Page.vue
jenow: We never show any trace in the admin console and I think if you want to show traces we should do that only in DEV mode and not show traces in production mode.

xbill82: True. Even if completely overkill, since we use this tool to manage one's own instances.
An still, the message in the toast is completely.

But I remember I found a Vuejs logger plugin that enabled to have logs being displayed under given conditions (like you said, dev mode or test - based on env vars).

jenow: That would be nice! I found this one which look promising https://www.npmjs.com/package/vuejs-logger

xbill82: :heavy_check_mark: Done!

PR #544, opened 2019-06-21
[HOTFIX] only fetch indexes and collections when adding or removing one

jenow:
## What does this PR do ?

Only refetch indexes and collections when one is added or removed.


src/components/Common/JsonEditor.vue
```diff
@@ -104,8 +105,10 @@ export default {
   methods: {
     getJson() {
       try {
-        return Object.freeze(JSON.parse(this.editor.getValue()))
+        // return Object.freeze(JSON.parse(this.editor.getValue()))
```
alexandrebouthinon: Did you forget to delete this line?
```suggestion
```

PR #549, opened 2019-06-25
Fix dependency security issue

jenow:
## What does this PR do ?

Update version of webpack-bundle-analyzer because github notices us of a securiy issue before the version 3.3.2.


xbill82:
BTW we should use this module to optimize the bundle size

Issue #547, opened 2019-06-24
cannot update a document where the identifier contains a space


barroudjo:
cannot update a document where the identifier contains a space

## Expected Behavior
Document should be updated when we click on update

## Current Behavior
Error message: An error occurred while trying to update the document: 
Not Found

## Possible Solution
The url for the update page is wrong, for the document with identifier "toto tata" the url is:
`http://console.kuzzle.io/#/data/institution/institution/update/toto%2520tata`
which contains a "%2520" instead of the "%20" that would be expected to describe the space (between "toto tata").
And indeed when I correct the url the updates happens as it should.

## Steps to Reproduce
1. create a document with a space in the identifier
2. try to update it
3.
4.

## Context (Environment)
Kuzzle version: 1.8.2 (but not relevant)
Node.js version: 12.4
Admin Console Version: online version, so latest

PR #567, opened 2019-09-04
Fix impossibility to edit documents with spaces in id (issue #547)

barroudjo:
## What does this PR do ?
Solves https://github.com/kuzzleio/kuzzle-admin-console/issues/547

Removed encodeURIComponent of params from push of new url to router, as it is already done by the router
 
  - Step 1 : create a document with a space in its id
  - Step 2 : click on the edit button
  - Step 3 :  check that the document is indeed opened

### Boyscout
In fact all encodeURIComponent in router push params should be removed (there are a few more than this one). But I don't know how to test there are no side effects...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10821,9 +10821,9 @@
       }
     },
     "node-sass": {
-      "version": "4.11.0",
-      "resolved": "https://registry.npmjs.org/node-sass/-/node-sass-4.11.0.tgz",
-      "integrity": "sha512-bHUdHTphgQJZaF1LASx0kAviPH7sGlcyNhWade4eVIpFp6tsn7SV8xNMTbsQFpEV9VXpnwTTnNYlfsZXgGgmkA==",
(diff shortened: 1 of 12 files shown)


xbill82:
Thanks for your contribution @barroudjo !
I don't think there are potential side effects in adding `encodeURIComponent` to the router params so we should clean them out. I'm adding this to my backlog.

barroudjo:
@xbill82 Tests are failing ? I didn't notice, where can I see which tests fail ?

xbill82:
@barroudjo we have a TravisCI instance running Cypress.io end-to-end tests

barroudjo:
> @barroudjo we have a TravisCI instance running Cypress.io end-to-end tests

@xbill82  where could I see the logs, so that I can fix the problem ? In the meantime I added commits that remove all encodeURIComponents and decodeURIComponents .

PR #585, opened 2019-10-21
Fix select list in roles and profiles

jenow:
## What does this PR do ?

Fix a bug where when we select a role or profile with the checkbox and delete it, this one would still be in memory of the internal selected documents so when selecting one other document after delete it would think it has 2 documents selected.

Also refresh the content after creating a new profile.

### DIFF
--- a/src/components/Security/Profiles/CommonList.vue
+++ b/src/components/Security/Profiles/CommonList.vue
@@ -127,13 +127,11 @@ export default {
   },
   methods: {
     async deleteProfiles(index, collection, ids) {
-      try {
-        await this.performDelete(index, collection, ids)
-        this.$set(this.selectedDocuments, this.selectedDocuments.splice(0, this.selectedDocuments.length))
-        this.fetchProfiles()
-      } catch (e) {
-        return Promise.reject(e)
-      }
+      await this.performDelete(index, collection, ids)
+      this.$set(
+        this.selectedDocuments,
+        this.selectedDocuments.splice(0, this.selectedDocuments.length))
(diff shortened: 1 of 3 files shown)


src/components/Security/Profiles/CommonList.vue
```diff
@@ -126,6 +126,15 @@ export default {
     )
   },
   methods: {
+    async deleteProfiles(index, collection, ids) {
+      try {
+        await this.performDelete(index, collection, ids)
+        this.$set(this.selectedDocuments, this.selectedDocuments.splice(0, this.selectedDocuments.length))
+        this.fetchProfiles()
+      } catch (e) {
+        return Promise.reject(e)
```
scottinet: (nitpicking) Throwing an error inside an async function is exactly like returning a rejected promise, so you can get rid of the whole `try..catch` statements.

src/components/Security/Roles/List.vue
```diff
@@ -119,6 +119,15 @@ export default {
     )
   },
   methods: {
+    async deleteRoles(index, collection, ids) {
+      try {
+        await this.performDelete(index, collection, ids)
+        this.$set(this.selectedDocuments, this.selectedDocuments.splice(0, this.selectedDocuments.length))
+        this.fetchRoles()
+      } catch (e) {
+        return Promise.reject(e)
```
scottinet: (nitpicking) Throwing an error inside an async function is exactly like returning a rejected promise, so you can get rid of the whole `try..catch` statements.

PR #594, opened 2019-12-16
Typescript support

jenow:
# What does this PR do ?

Add TS support.


src/directives/Materialize/m-select.directive.ts
```diff
@@ -5,12 +5,12 @@ export default {
...
-    let $el = $(el)
+    let $el: any = $(el)
     setTimeout(() => {
       $el.material_select()
-      $el.on('change', () => {
-        this.set(el.value)
-      })
+      // $el.on('change', () => {
+        // this.set(el.value)
+      // })
```
xbill82: Comment left accidentally

```suggestion
```

PR #606, opened 2020-01-17
[KZL-1433] 404 page

berthieresteban:
### What does this PR do ?

Use bootstrap card for the 404 page
Add '\*' path in router to display the 404 page for all non listed roads
Split 404 page in two components (1 for path '\*' and one in data folder for index/collection not found)
Clean home (wrapper class) to allow layout get full page height 
Use bootstrap in topBar, treeview, indexbranch

Update 404 cypress test to use data-cy instead of class


src/components/404.vue
```diff
@@ -1,15 +1,23 @@
...
-      <p>The index or collection that you requested doesn't exist.</p>
-      <p>
-        If you have a lot of indexes and/or collections please wait until those
-        are loaded in the left menu and this 404 will desappear.
-      </p>
-      <router-link class="404BackToHome-link" :to="{ name: 'Data' }">
-        Go back to the main page
-      </router-link>
+  <div class="Main">
+    <div class="fourOhFourPage">
```
xbill82: Why do we have this?
```
  Main
    fourOhFourPage
```
What is `Main` for?

src/components/Home.vue
```diff
@@ -6,16 +6,13 @@
...
-      <warning-header
-        v-if="!$store.direct.getters.auth.adminAlreadyExists"
-        :text="warningHeaderText"
-      />
-      <div class="wrapper">
-        <router-view />
-      </div>
-    </main>
-
+    <warning-header
```
xbill82: Should be replaced by `b-alert`

PR #607, opened 2020-01-22
[KZL-1447] App layout improvements

xbill82:
## What does this PR do ?
Cleans the application layout following an advice from Florens Verschelde. The main blocks of the application are now `display: flex` and we do not use the window scroll bar anymore. Instead, we force the scroll bars to appear, when needed, in the inner blocks (like, for instance, the sidebar and the main content block).

Check the Netlify build.

### Other changes
* Some tests have been fixed to use the new `data-cy` target selectors introduced in the previous works.
* Some error pages have been migrated

### Boyscout
* Some useless portions of code have been deleted

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -20,7 +20,7 @@
           v-if="!$store.direct.getters.kuzzle.currentEnvironmentId"
           class="App-disconnected"
         ></div>
-        <div v-else data-cy="App-connected" class="h-100">
+        <template v-else data-cy="App-connected" class="h-100">
           <div
             v-if="!$store.direct.getters.auth.isAuthenticated"
             class="App-loggedOut"
@@ -43,13 +43,13 @@
               />
             </div>
           </div>
-          <div v-else class="App-loggedIn h-100" data-cy="App-loggedIn">
+          <template v-else class="App-loggedIn h-100" data-cy="App-loggedIn">
             <router-view
               @environment::create="editEnvironment"
               @environment::delete="deleteEnvironment"
             />
-          </div>
-        </div>
+          </template>
+        </template>
       </template>
     </template>
 
@@ -61,7 +61,7 @@
     <modal-delete id="modal-env-delete" :environment-id="environmentId" />
     <modal-import id="modal-env-import" />
 
-    <toaster />
+    <!-- <toaster /> -->
   </div>
 </template>
 
@@ -78,7 +78,7 @@ import CreateEnvironmentPage from './components/Common/Environments/CreateEnviro
 import ModalCreateOrUpdate from './components/Common/Environments/ModalCreateOrUpdate'
 import ModalDelete from './components/Common/Environments/ModalDelete'
 import ModalImport from './components/Common/Environments/ModalImport'
-import Toaster from './components/Materialize/Toaster.vue'
+// import Toaster from './components/Materialize/Toaster.vue'
 
 // @TODO we'll have to import FA from global.scss one day...
 import '@fortawesome/fontawesome-free/css/all.css'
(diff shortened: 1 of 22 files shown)


src/App.vue
```diff
@@ -151,8 +149,6 @@ export default {
 
 <style lang="scss" scoped>
 .App {
-  height: 100vh;
-  overflow-x: hidden;
-  background-color: $page-background-color;
```
Njuelle: we need to keep the background color

tests/e2e/cypress/integration/search.spec.js
```diff
@@ -453,15 +450,15 @@ describe('Search', function() {
 
     cy.get('.RawFilter-submitBtn').click()
 
-    cy.get('.DocumentListItem').should(function($el) {
+    cy.get('.DocumentListItem').should(function ($el) {
       expect($el.first()).to.contain('Maret')
```
berthieresteban: Shouldn't be like line 385 ?
```suggestion
      expect($el.first()).to.contain('maret')
```

tests/e2e/cypress/integration/search.spec.js
```diff
@@ -453,15 +450,15 @@ describe('Search', function() {
...
       expect($el.last()).to.contain('Marchesini')
```
berthieresteban: Shouldn't be like line 386 ?
```suggestion
      expect($el.last()).to.contain('marchesini')
```

src/components/Common/MainMenu.vue
```diff
@@ -1,63 +1,60 @@
...
+      />
+    </b-navbar-brand>
+    <b-navbar-toggle target="nav-collapse" type="light"></b-navbar-toggle>
+    <b-collapse id="nav-collapse" is-nav>
+      <b-navbar-nav>
+        <b-nav-item
+          :active="
+            $route.path.match('/data')
+              ? $route.path.match('/data').length > 0
+              : false
```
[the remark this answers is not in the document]
xbill82: nope. The component wants a boolean prop

src/components/Common/MainMenu.vue
```diff
@@ -1,63 +1,60 @@
...
+          "
+          :to="{ name: 'Data' }"
+          >Data</b-nav-item
+        >
+        <b-nav-item
+          :active="
+            $route.path.match('/security')
+              ? $route.path.match('/security').length > 0
+              : false
```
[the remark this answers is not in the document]
xbill82: same here

tests/e2e/cypress/integration/search.spec.js
xbill82: No, because in the test at line 329, I changed the fixtures I inject (which, yeah, are not centralized in a proper fixture file, but are repeated in each test, sorry).
Actually, I could have avoided changing these tests. They fail because we ripped out all the Materialize CSS and the jQuery logic, so legacy dropdowns and other "complex" widgets aren't working anymore. Once we migrate the search UI, we'll have to entirely re-write the tests, so my changes are irrelevant at the time being.

tests/e2e/cypress/integration/search.spec.js
xbill82: same as previous

src/App.vue
xbill82: done!

src/components/Data/Layout.vue
```diff
@@ -108,19 +106,25 @@ export default {
...
 
-.content {
-  width: calc(100% - #{$sidebar-width});
-  margin-left: $sidebar-width;
+.DataLayout-contentWrapper {
+  flex-grow: 1;
+  height: 100%;
+  overflow: auto;
   padding: $content-gutter;
+  background-color: #f5f5f5;
```
Njuelle: we have a variable with this color already set

xbill82: ok

PR #609, opened 2020-01-27
[KZL-1434] Indexes bootstrap migration

xbill82:
## What does this PR do ?
Migrate the indexes list to BootstrapVue

Check the build visually


src/components/Data/Indexes/Page.vue
```diff
@@ -0,0 +1,187 @@
...
+      >
+        <i class="fa fa-plus-circle left" />
+        Create an index
+      </b-button>
+    </headline>
+
+    <list-not-allowed v-if="!canSearchIndex()" />
+    <template v-else>
+      <b-row class="mb-3">
+        <b-col sm="8" class="text-secondary">
```
Njuelle: this div is not vertically aligned with the input filter on the right.
```suggestion
        <b-col sm="8" class="text-secondary pt-2">
```

src/components/Data/Indexes/Page.vue
```diff
@@ -0,0 +1,187 @@
...
+          {{ tableItems.length }}
+          {{ tableItems.length === 1 ? 'index' : 'indexes' }}
+        </b-col>
+        <b-col sm="4">
+          <b-input-group>
+            <template v-slot:prepend>
+              <b-input-group-text>Filter</b-input-group-text>
+            </template>
+            <b-form-input v-model="filter"></b-form-input>
```
Njuelle: Maybe we can disable this input if there is no indexes

src/components/Data/Indexes/CreateIndexModal.vue
```diff
@@ -0,0 +1,86 @@
...
+      >
+        <b-form-input
+          data-cy="CreateIndexModal-name"
+          id="indexName"
+          required
+          type="text"
+          v-model="index"
+        ></b-form-input>
+      </b-form-group>
+      <b-alert
```
Njuelle: image

src/components/Data/Indexes/Page.vue
```diff
@@ -0,0 +1,187 @@
...
+        </template>
+        <template v-slot:cell(indexName)="indexName">
+          <router-link
+            :data-cy="`IndexesPage-name--${indexName.value}`"
+            :to="{
+              name: 'DataIndexSummary',
+              params: { index: indexName.value }
+            }"
+          >
+            {{ indexName.value }}
```
Njuelle: image

plus, In my opinion, it's not easy to know that you can click on the name of the index to see its contents without passing the cursor over it.

src/components/Data/Indexes/Page.vue
```diff
@@ -0,0 +1,187 @@
...
+          </router-link>
+        </template>
+        <template v-slot:cell(actions)="row">
+          <b-button
```
Njuelle: also we can add a create collection button here

berthieresteban: Not sure about that point

Njuelle: Why? Historically, the button existed in the previous version

xbill82: I agree with @berthieresteban : it looks like a dumb feature to me, but I also agree with @Njuelle : it was there before and taking it away may break some users' habits.
In the end, I feel it's not such an evil button, so I'd keep it.

PR #611, opened 2020-02-06
Migrate Create or Update collection to Bootstrap

xbill82:
## What does this PR do ?
Migrates Create or Update collection to Bootstrap

Build locally and check.


src/components/Data/Collections/CreateOrUpdate.vue
```diff
@@ -1,137 +1,232 @@
...
+              >
+              <br /><br />
+              For example:
+              <pre>
+{
+"age": { "type": "integer" },
+"name": { "type": "text" }
+}
+        </pre
+              >
```
berthieresteban:
```suggestion
              <pre>
{
  "age": { "type": "integer" },
  "name": { "type": "text" }
}
              </pre>
```

src/components/Data/Collections/CreateOrUpdate.vue
```diff
@@ -1,137 +1,232 @@
...
-    index: String,
-    headline: String
+    index: { type: String, required: true },
+    collection: String,
+    headline: String,
+    submitLabel: { type: String, default: 'OK' },
+    realtimeOnly: Boolean,
+    mapping: {
+      type: Object,
+      default: () => ({})
```
berthieresteban: I've nothing to say about that but I'm interested about why we could not use `default: {}`

xbill82: Actually, Vuejs forces the dev to do it this way.

PR #617, opened 2020-02-24
Simpler environment management

xbill82:
## What does this PR do ?
Refactors the environment management, fixing some small issues and preparing the ground for some bigger refactors.

- At the first page load, pay attention to the small logo and spinner at the center of the screen. They should be nicely centered.
- Create a valid connection (e.g. to your local kuzzle) and ensure the AC properly connects to it
- Create an invalid connection (with a random IP) and take a look at the error page it displays
- Switch between the two environments and ensure the AC smoothly switches from one to the other.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,92 +4,33 @@
     <meta charset="utf-8" />
     <title>Kuzzle Admin Console</title>
     <style>
-      .app-loading {
-        position: fixed;
-        top: 0;
-        left: 0;
-        bottom: 0;
-        right: 0;
-        padding-top: 50px;
-        height: 100%;
-        overflow: auto;
-      }
       html {
-        background-color: #f5f5f5;
+        /* We can avoid setting height:100vh on the grid container if we use height:100% 
+          on the root element and then all descendants up until the grid container. 
+          Using 100vh can be a simpler option, but the vh unit can be problematic on 
+          iOS and Android: https://nicolas-hoizey.com/articles/2015/02/18/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers/ 
+        */
+        height: 100%;
+        /* With our styles we should not have any cause for a scrollbar on the page itself,
+          so this is not strictly necessary, but to be safe we can prevent scrollbars on
+          the full page. 
+        */
+        overflow: hidden;
       }
-      .app-loading > div {
-        box-sizing: border-box;
-        font-family: Arial, sans-serif;
-        padding: 30px 40px;
-        width: 400px;
+
+      body {
+        height: 100%;
+      }
+
+      .center-screen {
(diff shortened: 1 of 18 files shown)


berthieresteban:
Create ''local' environment
Create a fake 'test' environment
Switch between
On 'test' environment we are on the unable to connect to kuzzle server with the spinner
Delete 'test' env
We can't delete 'test' env, the modal stay on the screen after validating the environment name
The selectbox behind the modal switch to 'null'
The console display some errors:
  - `Error: Cannot connect to a null environment. To reset connection, use the RESET mutation.`
  - `Uncaught (in promise) Error: Unable to execute request: not connected to a Kuzzle server.`

berthieresteban:
Using a fake 'test' env
On trying to connect page
Update the environment name
Get a non explicit line  `[Object Event]`
image

berthieresteban:
The environment name in the select must be cut 
image

berthieresteban:
When trying to connect to a fake 'test' environment
Update the environment name of the 'local' environment
The select switch to the 'local' environment and connect us and push us on the index page while we were trying to connect to the 'test' environment

berthieresteban:
Same as above on the index page
When connected to the 'local' environment
Update the 'test' environment name change the current connection and push us on trying to connect to test:7512 page

berthieresteban:
When connected to the current 'local' environment,
In the switch we can access the current environment
When trying to delete it
Push on the connection page (login as anonymous or create an admin)
And the modal still present
image

berthieresteban:
environment name on index page should be cut
image

berthieresteban:
Same here we should cut the enrivonment name 
image

berthieresteban:
Username should be cut too 
image

xbill82:
Also, the "Export environments" button doesn't work for me. Do you confirm that?

berthieresteban:
Run kuzzle stack with kourou
Connect to AC on localhost
Try to create first admin
Get an error (console + toast)
Create admin button stay disabled until we reload or click on the current environment in the select env
image

berthieresteban:
> Also, the "Export environments" button doesn't work for me. Do you confirm that?

Yes it doesn't work for me too

xbill82:
Hi @berthieresteban , concerning the errors below

> The console display some errors:
>
>    Error: Cannot connect to a null environment. To reset connection, use the RESET mutation.
>    Uncaught (in promise) Error: Unable to execute request: not connected to a Kuzzle server.

I think we should fix this in the #618 because it will be way easier to do it (the refactored code is much cleaner)

xbill82:
Yo @berthieresteban I've solved the issues that I've tagged with the :+1: emoji. The ones that are tagged with the :rocket: emoji should be solved in the #618 PR.

PR #618, opened 2020-02-25
Refactor de la muerte

xbill82:
## What does this PR do ?
Something radically out of scope but that fixes a behavior everybody was complaining about.

Makes the AC properly handle the "Connecting...", "Connected", "Temporarily offline" states. Now, whenever Kuzzle goes down, the AC just shows a toast instead of changing the current page.

To implement this, a whole lot of stuff has been refactored. A whole lot of junky code has been cleaned up and it feels damn good.

* Launch a Kuzzle instance
* Connect the AC to it
* Stop the Kuzzle instance
* Verify that the AC displays a toast
* Relaunch the Kuzzle instance
* Verify that the toast is gone

:warning: All the tests of the following specs must be :heavy_check_mark: 
* 404 page
* Collection management
* Document list
* Document update/replace
* Environments
* Indexes
* Login
* Search

### Other changes and Boyscout
- [x] Improve the handling of JSON parsing errors on environment load: a toast should notify that the environments are malformed and dump the content of the LocalStorage item to console.
- [x] Handle the case where no current environment is set (currently defaults to first available one)
- [x] Properly handle delete environments, both when still env left and when no env left.
- [x] Nice and smooth environment import with proper error messages
- [x] Nice and smooth entironment switch
- [x] Improve the handling of an environment update: if the URL of the Kuzzle server has changed, the connection should be reset.
- [x] Proper first admin check
- [x] Fix tests
- [x] Connection spinner on first page load
- [x] Token expired check


berthieresteban:
I just tested it and it's very cool :+1: 

When the toast is displayed, we can access all other features. I tried to create an index while no kuzzle was running and I get an error not very explicit.
Maybe we could add an overlay or something to avoid user continue using AC features while the toast is displayed ?

xbill82:
@berthieresteban The idea of the toast would rather be to allow the user to continue using _the current page_ without connectivity. Essentially, the aim is not losing the form data that have been typed in the fields, or even keep on typing without connectivity, so the overlay seems a bit too much to me.

But indeed, we should 
* create reliable error workflows, with meaningful messages,
* prevent the user to click on links that would lead do a new page, which would not be able to display itself, since no `checkToken` action would be possible,
* prevent the user to trigger actions that involve interactions with the server (like, e.g. creating a new index).

For this purpose, I was thinking about some sort of "connection-aware buttons", that would disable themselves when the server is not reachable. Does this make sense to you?

berthieresteban:
I think there is two subjects here that might be treat in two times:
  - Keep the user on the current page while loosing the connection to allow him `not losing the form data that have been typed in the fields`
  - And allow the user to `keep on typing without connectivity`

The ideas you propose in the previous message are very great and fit very well with both of the two subjects.

But IMHO, the second subject could be a new feature of the admin console and might take a more important time than the first. 

That's why I think adding an overlay could be a good compromise because it allow us to treat the first subject avoiding some bugs and we could treat the second subject after the end of the redesigned phase.

xbill82:
@berthieresteban I understand your statement and I think it's a very good point. But I am having a hard time accepting this idea of the overlay because it kind of brings the whole thing back to where we are now.
The current state of the AC is "You lose the connectivity === you lose the page", which is unacceptable. The overlay would mean "You lose the connectivity === you stop touching stuff", which is not as bad, but still... It's shitty.
What I propose is: let's stick with this bold idea of just providing something really good without rushing on the release of this feature.
Let's take the time to test everything and write Cypress tests for everything.
What I'd like to achieve in the short term is *meaningful error messages* which is _always_ a good thing.

What do you think?

berthieresteban:
***after a wild discussion***

I agree with you and the only important difference between our propal is that we allow user found some bug with yours BUT since we will not release the AC soon and as you said we could ask some beta-tester, your solution looks betters because those BT will be able to spot more bugs.

berthieresteban:
Create ''local' environment
Create a fake 'test' environment
Switch between
On 'test' environment we are on the unable to connect to kuzzle server with the spinner
Delete 'test' env
We can't delete 'test' env, the modal stay on the screen after validating the environment name
The selectbox behind the modal switch to 'null'
The console display some errors:
  - `Error: Cannot connect to a null environment. To reset connection, use the RESET mutation.`
  - `Uncaught (in promise) Error: Unable to execute request: not connected to a Kuzzle server.`

berthieresteban:
Using a fake 'test' env
On trying to connect page
Update the environment name
Get a non explicit line  `[Object Event]`
image

berthieresteban:
When trying to connect to a fake 'test' environment
Update the environment name of the 'local' environment
The select switch to the 'local' environment and connect us and push us on the index page while we were trying to connect to the 'test' environment

xbill82:
Unable to edit an existing environment from the Login screen

xbill82:
Import environments doesn't successfully switch to the imported env

xbill82:
Switching from an env to anorther doesn't refresh the index list

berthieresteban:
Capture du 2020-03-05 14-00-12
I'm not sure but we might check the `type of the property` and the `verb condition` to display a better error like `You should use the match condition with a string property` or `Number property and match condition are incompatible`.

berthieresteban:
Capture du 2020-03-05 13-58-56
We will probably need to crop the displayed json in list view.

berthieresteban:
Capture du 2020-03-05 13-58-40
I can't click on the `i` icon to get the tooltip displayed

berthieresteban:
On create collection page, click on cancel button display a blank page

berthieresteban:
Create document, cancel, we probably should display a spinner instead of `this collection is empty` during the loading

src/components/Data/Collections/CreateOrUpdate.vue
```diff
@@ -183,6 +183,16 @@ export default {
...
+    cancel() {
+      if (this.$router._prevTransition && this.$router._prevTransition.to) {
+        this.$router.push(this.$router._prevTransition.to)
+      } else {
+        this.$router.push({
+          name: 'Indexes',
+          params: { index: this.index }
+        })
+      }
+    },
```
berthieresteban: On create collection page, click on cancel button display a blank page

src/components/Data/Documents/Page.vue
```diff
@@ -29,84 +32,96 @@
...
+            <b-col cols="12">
+              <template v-if="isCollectionEmpty">
+                <realtime-only-empty-state
+                  v-if="isRealtimeCollection"
+                  :index="index"
+                  :collection="collection"
+                />
+                <empty-state v-else :index="index" :collection="collection" />
+              </template>
+              <template v-if="!isCollectionEmpty">
```
berthieresteban:
```suggestion
              <template v-else>
```

xbill82:
Gets back to login page on reload... can you confirm @berthieresteban ?

PR #619, opened 2020-03-11
fix tooltip

berthieresteban:
## What does this PR do ?
On the column view, for arrays values an icon is displayed with a tooltip to tells why we can't display the value. 
This tooltip was only displayed when hovering the dot of the 'i' icon.
This pr fix that by using bootstrap vue tooltip (wich is prettier...)

  - Run a kuzzle
  - Create a collection with a field containing an array of what you want.
  - Create a document with an array in that field
  - Go to admin console column view and add this field
  - hover the tooltip 

### Other changes
/

### Boyscout
/

/
before: 
Capture d’écran de 2020-03-11 15-25-44

after:
Capture d’écran de 2020-03-11 15-25-24

### DIFF
--- a/src/components/Data/Documents/Views/Column.vue
+++ b/src/components/Data/Documents/Views/Column.vue
@@ -186,12 +186,14 @@
                 :id="`tooltip-target-${data.item.id}-${data.field.key}`"
                 v-if="data.value.array === true"
               >
-                <i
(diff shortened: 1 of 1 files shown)


xbill82:
Screenshot from 2020-03-17 16-55-17

Fix in progress...

PR #621, opened 2020-03-13
format message wrong query

berthieresteban:
## What does this PR do ?
The toast of the error when a query was wrong displayed the entire error message. 
This Pr just makes it understandable

  - Step 1 : Run Kuzzle and go on AC
  - Step 2 : If needed, create a collection and a document with a text field for example
  - Step 3 : in advanced mode query search use a maladjusted verb (ex: match for number or equal for text)
  - Step 4 : Check the toast displayed on the right bottom
  ...

### Other changes
/

### Boyscout
/

Capture du 2020-03-05 14-00-12
Capture du 2020-03-13 09-37-56

### DIFF
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -484,6 +484,19 @@ export default {
         this.$log.error(e)
         if (e.status === 412) {
           this.indexOrCollectionNotFound = true
+        } else if (e.message.includes('failed to create query')) {
+          this.$bvToast.toast(
+            'Your query is ill-formed. The complete error has been dumped to the console.',
+            {
+              title:
+                'Ooops! Something went wrong while fetching the documents.',
+              variant: 'warning',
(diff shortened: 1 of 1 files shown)


src/components/Data/Documents/Page.vue
```diff
@@ -484,6 +484,19 @@ export default {
         this.$log.error(e)
         if (e.status === 412) {
           this.indexOrCollectionNotFound = true
+        } else if (e.message.includes('failed to create query')) {
```
jenow:
```suggestion
        } else if (e.id === 'services.storage.unexpected_bad_request') {
```

xbill82: @jenow not available yet since we're still on the SDK 7.0.0-beta (which is shit, we should bump ASAP).
Once we bumped, your remark must be extended to all the cases where we perform pattern matching on the SDK error messages.

PR #622, opened 2020-03-16
[KZL-1441] realtime view

xbill82:
## What does this PR do ?
Redesigns the Realtime view for a collection.

* Run the tests and ensure that _all the tests except from `Users` are green_
* play with it on Netlify

## Other changes
* Rewritten the API of the `JSONEditor` component to make it more usable and reliable


src/components/Data/Collections/Watch.vue
```diff
@@ -427,111 +360,69 @@ export default {
...
+  async destroyed() {
+    this.reset()
+    if (this.room) {
+      await this.$kuzzle.realtime.unsubscribe(this.room)
     }
   }
 }
 </script>
 
 <style rel="stylesheet/scss" lang="scss">
```
Njuelle: Why not scoped style here? 
if needed it is possible to put 2 styles tags, one scoped and the other not, which allows to better separate the styles and to be sure to have a minimum of css override

src/components/Data/Realtime/SubscriptionControls.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="valign-center">
     <div class="col s5 m5 l4">
-      <button
+      <!-- <button
```
Njuelle: maybe you forgot to clean here, or you don't want to lose this precious code ? ;)

xbill82: I'd rather delete the whole component, since we use it no more...

PR #629, opened 2020-03-25
Display human readable date for string timestamp

Aschen:
## What does this PR do ?

Elasticsearch can store date timestamp either with integer or with string.  
Actually, we display an human readable date only for the integer timestamp.  

This PR allow to display an human readable date also with string timestamp.

image

### DIFF
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -642,12 +642,8 @@ export default {
 
       const changeField = document => {
         for (const [field, value] of Object.entries(document)) {
-          if (dateFields.includes(field) && Number.isInteger(value)) {
-            const date =
-              `${value}`.length === 13
-                ? new Date(value)
-                : new Date(value * 1000)
-
+          if (dateFields.includes(field)) {
+            const date = dateFromTimestamp(value)
             document[field] += ` (${date.toUTCString()})`
           } else if (value && typeof value === 'object') {
             changeField(value)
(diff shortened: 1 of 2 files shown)


xbill82:
oh! yeah I implemented it for the Meta in #624 but didn't think to do it for the content! We should harmonize the format, btw

src/components/Data/Documents/Page.vue
```diff
@@ -661,6 +657,35 @@ export default {
...
 }
+
+function dateFromTimestamp(value) {
+  let timestamp
+
+  if (typeof value === 'string') {
+    timestamp = parseInt(value, 10)
+
+    if (isNaN(timestamp)) {
+      return null
```
xbill82: What about displaying something instead of null?
```suggestion
      return `Invalid Date value (${value})`
```

src/components/Data/Documents/Page.vue
```diff
@@ -661,6 +657,35 @@ export default {
...
+
+  const length = `${timestamp}`.length
+
+  let date
+  if (length === 10) {
+    date = new Date(timestamp * 1000)
+  } else if (length === 13) {
+    date = new Date(timestamp)
+  } else {
+    date = null
```
xbill82:
```suggestion
    return `Invalid Date value (${value})`
```

Aschen:
Good idea @xbill82  :+1:

PR #630, opened 2020-03-25
[KZL-1484] Refactor list collection indexes

xbill82:
## What does this PR do ?
Enhances the experience of loading the index and collection list. Now indexes are loaded first (thus rendered), then collections for each index are loaded afterwards. Some glitches still persist in the Treeview but the overall bootstrap process is better.

Connect the AC to a Kuzzle with a lot of indexes and collections and see if the bootstrap is smooth.

### Other changes
* Applied `scroll` to document searches like #632

### TODO
- [x] Loader when opening the collection list for an index


src/vuex/modules/index/store.ts
```diff
@@ -104,33 +101,55 @@ const actions = createActions({
...
-        getRealtimeCollectionFromStorage(index)
-      )
-      collections = dedupeRealtimeCollections(collections)
-      indexesAndCollections[index] = collections
-      commit.receiveIndexesCollections(indexesAndCollections || {})
-    }
-    commit.setLoading(false)
+    await dispatch.listIndexes()
+    Object.keys(state.indexesAndCollections).forEach(async index => {
+      await dispatch.listCollectionsForIndex(index)
```
Aschen, line 151: Awaiting a promise inside a forEach won't really wait for the promise to complete before processing the next item on the loop.  
This method will return before each promise is fulfilled, I don' know if this is the intended behavior.

src/components/Data/Collections/List.vue
```diff
@@ -26,131 +26,134 @@
...
+            <p
+              v-if="canCreateCollection(index)"
+              class="text-secondary text-center"
+            >
+              You can create the collection by hitting the button above.
+            </p>
+          </template>
+          <template v-slot:emptyfiltered>
+            <h4 class="text-secondary text-center">
+              There is no collection matching your filter.
```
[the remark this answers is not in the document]
xbill82: nope

src/vuex/modules/index/store.ts
xbill82: Thanks! I forgot to delete the `await`

PR #631, opened 2020-03-25
Display complete index and collection names

Aschen:
## What does this PR do ?

Display complete index name on the `CollectionList` page 

image

### Other changes
 
 - add a home icon on breadcrumb
 - auto use `ssl` when port is 443


Aschen:
So we can merge this one ? @xbill82

PR #634, opened 2020-03-30
[KZL-1439] User list

xbill82:
## What does this PR do ?
Migrates the user list to Bootstrap V4 and implements the necessary tests.

:warning: @berthieresteban initially started this work and he began to migrate the Update User screen, but it will be finished in a subsequent PR. Please do not review it.

Play around with the users, take a look at the e2e tests.


Aschen:
Nice work :+1:  
I came across few bug and have few potential suggestions:

#### The `aggregations` field should not be present on the user card
image

#### The metadata are not correctly displayed
I saw 
image
But my user have an `author`, a `createdAt`, an `updater` and a `updatedAt`

#### Local strategy username
It could be useful to display the `username` of the local strategy in the user card but maybe it can be done in a separate PR

#### Profile selection
On some project we can have 50-60 different profiles and it's difficult because they are not sorted alphabetically. 
Can you at least sort them? And the best UX would be to able to input text to filter the select box

#### User ID generation
Instead of having a checkbox `auto-generate`, we could just generate an ID if the user don't fill the input

#### User custom content
The mapping is displayed on a single line
image

#### Breadcrumbs
The breadcrumbs for the user section is not present on the creation form
image

#### Search bar
I don't understand exactly how the basic search is working, are you also searching on user profile ?
For example users may want to list users with the `admin` profile

#### User mapping

User mapping edition seems to be broken 
image

src/components/Security/Users/List.vue
```diff
@@ -0,0 +1,396 @@
...
+      this.deleteModalIsLoading = true
+      try {
+        await performDeleteUsers(
+          this.index,
+          this.collection,
+          this.candidatesForDeletion
+        )
+        this.$bvModal.hide('modal-delete-users')
+        this.deleteModalIsLoading = false
+        this.fetchDocuments()
```
Aschen, line 317: This call is not awaited nor the promise rejection catched

src/components/Security/Users/Steps/UserProfileList.vue
```diff
@@ -48,29 +47,29 @@
...
     }
   },
   computed: {
     availableProfiles() {
-      return this.profileList.filter(profile => {
-        return this.addedProfiles.indexOf(profile.id) === -1
-      })
+      return this.profileList
+        .filter(profile => {
+          return this.addedProfiles.indexOf(profile.id) === -1
```
Aschen, line 70: Nitpicking: prefer `Array.includes`

```suggestion
          return !this.addedProfiles.includes(profile.id)
```

src/components/Security/Users/UserItem.vue
```diff
@@ -1,96 +1,123 @@
...
-    </label>
-    <!-- The following anchor will go to the user details page -->
-    <label class="UserItem-title item-title">
-      <a @click="toggleCollapse">{{ document.id }}</a>
-      <div class="UserItem-profileList">
-        <div
+  <b-container fluid data-cy="UserItem">
+    <b-row align-h="between" no-gutters>
+      <b-col cols="10" class="py-1">
+        <!-- <div class="UserItem" :class="{ collapsed: collapsed }"> -->
```
Aschen, line 5: Leftover comments

src/components/Security/Users/UserItem.vue
```diff
@@ -1,96 +1,123 @@
...
+          class="d-inline-block align-middle"
+          type="checkbox"
+          value="true"
+          unchecked-value="false"
+          v-model="checked"
+          :data-cy="`UserListItem-checkbox--${document.id}`"
+          :id="checkboxId"
+          @change="notifyCheckboxClick"
+        />
+        <!-- The following anchor will go to the user details page -->
```
Aschen, line 26: Leftover comments

test/e2e/cypress/integration/users.spec.js
```diff
@@ -77,25 +256,53 @@ describe.skip('Users', function() {
...
-    cy.get('.CommonList').should('not.contain', kuid)
+    cy.get(
+      '[data-cy="UserManagement-pagination"] .page-link[aria-posinset="2"]'
+    ).click()
+    cy.get('[data-cy=UserItem]').should('have.length', 4)
+    cy.url().should('contain', 'from=10')
   })
 
-  it('updates the user mapping successfully', function() {
+  it.skip('updates the user mapping successfully', function() {
```
Aschen, line 305: Is this intentional?

src/components/Security/Users/List.vue
```diff
@@ -0,0 +1,396 @@
...
+        return false
+      }
+
+      return this.selectedDocuments.length === this.documents.length
+    },
+    paginationFrom() {
+      return parseInt(this.currentFilter.from) || 0
+    },
+    paginationSize() {
+      return parseInt(this.currentFilter.size) || 10
```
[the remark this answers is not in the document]
xbill82: This may be a good idea but we display 10 items by default in all the lists of the app: wouldn't it be weird to have one that displays 20?

src/components/Security/Users/List.vue
xbill82: Yes, it's because `fetchDocuments` already wraps sensitive code in a `try...catch` block and displays a toaster in case of error

test/e2e/cypress/integration/users.spec.js
xbill82: ops!

xbill82:
In response to @Aschen's big comment

*Aggregations, metadata & credentials*
I deleted aggregations and metadata and just showed the `credentials` object

*Profile selection*
Yeah, it's a real bummer... I'd like this to be implemented in a separate UX-enhancement PR.

*User creation remarks*
As stated in the PR description, this PR doesn't implement user creation, please refer to #636

src/components/Security/Users/List.vue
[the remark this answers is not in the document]
Aschen: The size of users object are roughly always the same but documents can differ and be very huge.
I don't think it's a technical problem.

About the UX, securities are displayed in a different page so I'm not choked if the list behavior is different.

src/components/Security/Users/List.vue
Aschen: Yes but since you are not awaiting this call, the `fetchDocuments` method will not throw on error but just return a rejected promise which end up unhandled

Consider this code
```
async function fetchDocuments(b) {
  if (b) {
    throw new Error()
  }
}

async function run () {
  try {
    fetchDocuments(true)
  }
  catch (error) {
    console.error(error)
  }
}

run()
```
This will not be catched by the try catch and you will end up with a `unhandledRejection` error

Aschen:
Missing breadcrumb on user edition

image
_should be security > users > admin_

User mapping edition is broken
image
_The breadcrumb is also missing _

xbill82:
@Aschen as specified in a [previous comment](https://github.com/kuzzleio/kuzzle-admin-console/pull/634#issuecomment-613371746) AND in the description of this PR, **the edit user feature is implemented in a subsequent PR** (#636). Please, **read**.

Aschen:
> @Aschen as specified in a [previous comment](https://github.com/kuzzleio/kuzzle-admin-console/pull/634#issuecomment-613371746) AND in the description of this PR, **the edit user feature is implemented in a subsequent PR** (#636).

Ok :+1: The breadcrumb is still missing on the other PR, I will add a review

PR #635, opened 2020-04-06
Fix collection list max size

Aschen:
## What does this PR do ?

Collection list method is paginated with a size of 10 by default so if there is more than 10 collections inside an index, they won't be displayed.  

We will remove the collection pagination from Kuzzle but to ensure that the Admin Console will work with previous version I put a very high default number


xbill82:
Doesn't `size: 0` work?

xbill82:
Netlify fails build while trying to DL Node v6 :point_right: need to change (or delete?) `.nvmrc`

xbill82:
Before merging we should see why Netlify shows the `Error: Cannot find module 'shelljs/global'`

xbill82:
OK, I found where the error comes from. Lately, I have set the `$NODE_ENV` variable to `production` in the Netlify build environment, so that Cypress could be specified as a dev-dependency thus not installed by Netlify. Unfortunately, all the important dependencies in the `3-dev` branch are expressed as dev-dependencies. It's fixed.

PR #636, opened 2020-04-06
[KZL-1439] Create or update user

xbill82:
:warning: Depends on #634 

## What does this PR do ?
Migrates to Bootstrap v4 the create user page and the update user page (which use the same component).

Create a user with some custom data and then update its profiles, credentials and custom data.

### Other changes
* Now showing credentials in the User list


Njuelle:
Overflow on profiles badges
image

Njuelle:
all the badges of the profiles are not displayed when we consult the list of users (whereas I have 20 badges on this user)
image

Njuelle:
When I create a user and I don't check the "auto-generate" box and leave the KUUID field empty, the user is still created with an auto-generated KUUID. it might be better to prevent the creation of the user and maybe show an error toast

src/components/Security/Users/CreateOrUpdate.vue
```diff
@@ -0,0 +1,339 @@
...
+                >Cancel</b-button
+              >
+              <b-button
+                class="m-1"
+                data-cy="UserUpdate-submit"
+                type="submit"
+                variant="primary"
+                @click.prevent="submit"
+              >
+                <span v-if="!!id">Save</span>
```
Njuelle, line 73:
```suggestion
                <span v-if="id">Save</span>
```

.vscode/toast.code-snippets
```diff
@@ -0,0 +1,25 @@
+{
```
Njuelle, line 1: smart ;)

src/components/Security/Users/DeleteModal.vue
```diff
@@ -0,0 +1,59 @@
...
+
+export default {
+  name: 'UserDeleteModal',
+  components: {},
+  props: {
+    id: {
+      type: String,
+      required: true
+    },
+    candidatesForDeletion: Array,
```
Njuelle, line 52: NIT: add a default value for not required props, or maybe they need to be required

.vscode/toast.code-snippets
Njuelle: But idk if we have to commit vscode snippets ?

src/components/Security/Users/Page.vue
```diff
@@ -106,3 +104,17 @@ export default {
   }
 }
 </script>
+
+<style lang="scss" scoped>
+::v-deep .usersDropdown {
```
Njuelle, line 109: wow nice :) 
I didn't know this `v-deep` tricks

src/components/Security/Users/Steps/CustomData.vue
```diff
@@ -1,87 +1,61 @@
...
-      </div>
+        <b-col>
+          <h3>Mapping</h3>
+          <pre
+            v-json-formatter="{
+              content: mapping,
+              open: true
+            }"
+          />
+          <!--<json-editor
```
Njuelle, line 30: remove useless old commented code

src/components/Security/Users/Steps/UserProfileList.vue
```diff
@@ -1,76 +1,81 @@
...
     }
   },
   computed: {
     availableProfiles() {
-      return this.profileList.filter(profile => {
-        return this.addedProfiles.indexOf(profile.id) === -1
-      })
+      return this.profileList
+        .filter(profile => {
+          return this.addedProfiles.indexOf(profile.id) === -1
```
Njuelle, line 76:
```suggestion
          return this.addedProfiles.includes(profile.id)
```

src/components/Security/Users/UserItem.vue
```diff
@@ -1,96 +1,123 @@
...
-    </label>
-    <!-- The following anchor will go to the user details page -->
-    <label class="UserItem-title item-title">
-      <a @click="toggleCollapse">{{ document.id }}</a>
-      <div class="UserItem-profileList">
-        <div
+  <b-container fluid data-cy="UserItem">
+    <b-row align-h="between" no-gutters>
+      <b-col cols="10" class="py-1">
+        <!-- <div class="UserItem" :class="{ collapsed: collapsed }"> -->
```
Njuelle, line 5: useless old commented code

src/components/Security/Users/UserItem.vue
```diff
@@ -1,96 +1,123 @@
...
+          type="checkbox"
+          value="true"
+          unchecked-value="false"
+          v-model="checked"
+          :data-cy="`UserListItem-checkbox--${document.id}`"
+          :id="checkboxId"
+          @change="notifyCheckboxClick"
+        />
+        <!-- The following anchor will go to the user details page -->
+        <!-- <label class="UserItem-title item-title"> -->
```
Njuelle, line 28: useless old commented code

.vscode/toast.code-snippets
xbill82: I'd rather do it since they're pretty handy and don't bother the devs that don't use vscode

src/components/Security/Users/DeleteModal.vue
xbill82: Why? if no default value is specified, the implicit is `null`, which is fine here, isn't it?

src/components/Security/Users/Page.vue
xbill82: This is something @berthieresteban did and I'm not sure how it works either

berthieresteban: v-deep is just a way to unscoped the following class
Just like if we add another `<style>` block without scoped and add the `usersDropdown` class in  it

src/components/Security/Users/Steps/UserProfileList.vue
```diff
@@ -86,6 +91,7 @@ export default {
     },
     onProfileSelected(profile) {
       this.$emit('selected-profile', profile)
+      this.selectedProfiled = 0
```
berthieresteban, line 94: When adding the last profile of the list, there is no more profiles to add so the selectedProfile changed and that trigger this function and add to the two level parent an undefined profile badge

```suggestion
    onProfileSelected(profile) {
      if (!profile) {
        return;
      }
      this.$emit('selected-profile', profile)
      this.selectedProfiled = 0
```

src/components/Security/Users/CreateOrUpdate.vue
```diff
@@ -0,0 +1,339 @@
...
+            this.kuid,
+            {
+              content: {
+                profileIds: this.addedProfiles,
+                ...JSON.parse(this.customContentValue)
+              },
+              credentials: this.credentials
+            },
+            {
+              refresh: 'wait_for'
```
Aschen, line 237: The security route default value for `refresh` is already `wait_for`

src/components/Security/Users/CreateOrUpdate.vue
```diff
@@ -0,0 +1,339 @@
...
+      try {
+        if (this.id) {
+          await this.$kuzzle.security.replaceUser(
+            this.kuid,
+            {
+              profileIds: this.addedProfiles,
+              ...JSON.parse(this.customContentValue)
+            },
+            {
+              refresh: 'wait_for'
```
Aschen, line 201: The security route default value for `refresh` is already `wait_for`

PR #637, opened 2020-04-06
[KZL-1439] Update user mapping

xbill82:
:warning: Depends on #636 

## What does this PR do ?
Migrates the Update User Mapping page to Bootstrap v4

Add some properties to the user mapping and ensure they have been correctly saved.


src/components/Security/Users/EditCustomMapping.vue
```diff
@@ -47,50 +85,72 @@
  */
 
 import Headline from '../../Materialize/Headline'
-import Mapping from './Steps/Mapping'
+import JsonEditor from '../../Common/JsonEditor'
 import {
   getMappingUsers,
   updateMappingUsers
 } from '../../../services/kuzzleWrapper'
+import { omit } from 'lodash'
```
rolljee, line 93: To avoid sending the whole lodash package to the client, it's better practice to import it as follow
`import omit from 'lodash/omit'`

PR #638, opened 2020-04-07
[KZL-1439] Profile list

xbill82:
:warning: Depends on #637 

## What does this PR do ?
Migrates the profile list to Bootstrap v4

Filter profiles and delete profiles


src/components/Security/Profiles/ProfileItem.vue
```diff
@@ -76,7 +99,7 @@ import title from '../../../directives/title.directive'
 export default {
   name: 'ProfileItem',
   components: {
-    Dropdown
+    // Dropdown
```
rolljee, line 102: Commented line should be removed

src/components/Security/Users/UserItem.vue
```diff
@@ -1,96 +1,123 @@
...
-      <a @click="toggleCollapse">{{ document.id }}</a>
-      <div class="UserItem-profileList">
-        <div
+  <b-container fluid data-cy="UserItem">
+    <b-row align-h="between" no-gutters>
+      <b-col cols="10" class="py-1">
+        <!-- <div class="UserItem" :class="{ collapsed: collapsed }"> -->
+        <!-- class="UserItem-toggle fa fa-caret-down item-toggle" -->
+        <i
+          aria-hidden="true"
```
rolljee, line 8: Ditto aria-hidden

src/components/Security/Profiles/ProfileItem.vue
```diff
@@ -1,71 +1,94 @@
...
-              title: 'You are not allowed to delete this profile'
-            }"
-            :class="{ disabled: !canDeleteProfile() }"
-            @click="deleteDocument(document.id)"
-            >Delete</a
+  <b-container fluid data-cy="ProfileItem">
+    <b-row align-h="between" no-gutters>
+      <b-col cols="10" class="py-1">
+        <i
+          aria-hidden="true"
```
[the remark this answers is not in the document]
xbill82: Oh great! I was totally unaware of these `aria` attributes, dunno where they come from. Thanks!

src/components/Security/Profiles/Filters.vue
```diff
@@ -1,79 +1,135 @@
...
+          >
+            Reset
+          </b-button>
+          <b-button
+            type="submit"
+            variant="primary"
+            @click.prevent="submitSearch"
+          >
+            {{ labelSearchButton }}
+          </b-button>
```
berthieresteban, line 66: I'm not sure about the necessity of this button. 
When we add a role in the selector, the list is automatically updated. 
This search button just use the router to push on the same page with the same filters so that trigger a `navigationDuplicated` error.
If we decided to keep that button to refresh the list displayed, we might need to change the way the list works (using the router).
But IMHO, it doesn't make sens because the list is always updated with the last role filters we setted.

xbill82: OK let's take it away

PR #642, opened 2020-04-21
[KZL-1439] Role list

xbill82:
Migrates the Role list to Bootstrap V4

### To Do
- [x] Layout
- [x] Role item appearence and collapse
- [x] Filters appearence
- [x] Filters behavior
- [x] Delete role(s) modal
- [x] Pagination
- [x] Tests

### DIFF
--- a/src/components/Security/Roles/Page.vue
+++ b/src/components/Security/Roles/Page.vue
@@ -1,6 +1,20 @@
 <template>
-  <div class="RolesManagement">
-    <headline title="Role Management" />
+  <b-container class="RolesManagement">
+    <b-row>
+      <b-col cols="8">
+        <headline>Roles</headline>
+      </b-col>
+      <b-col class="text-right mt-3">
+        <b-button
+          class="mr-2"
+          data-cy="RolesManagement-createBtn"
+          variant="primary"
+          :disabled="!canCreateRole()"
+          :to="{ name: 'SecurityRolesCreate' }"
+          >Create Role</b-button
+        >
+      </b-col>
+    </b-row>
 
     <list-not-allowed v-if="!canSearchRole()" />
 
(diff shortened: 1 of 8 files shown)


berthieresteban:
Great job ! 
Here we must click on search but we removed that button in profiles view.
I'm not sure what way is the best but we probably should have the same workflow for the filters and search in roles/profiles/users.

Another point, the searchRoles work the same way as searchProfile with an OR.
I think we should do something about that.

xbill82:
@berthieresteban actually, we should take out the Search button and trigger the new search each time a controller is added to the searchbar

src/components/Security/Roles/Page.vue
```diff
@@ -1,6 +1,20 @@
 <template>
-  <div class="RolesManagement">
-    <headline title="Role Management" />
+  <b-container class="RolesManagement">
+    <b-row>
+      <b-col cols="8">
```
rolljee, line 4: I was wondering if when specified in bs4 cols must be equals to 12 ? (wont block for that anyway)

src/components/Security/Roles/RoleItem.vue
```diff
@@ -1,81 +1,82 @@
 <template>
-  <div class="RoleItem" :class="{ collapsed: collapsed }">
-    <i
-      class="RoleItem-toggle fa fa-caret-down item-toggle"
-      aria-hidden="true"
-      @click="toggleCollapse()"
-    />
+  <b-container fluid data-cy="RoleItem">
+    <b-row align-h="between" no-gutters>
+      <b-col cols="10" class="py-1 vertical-align">
```
rolljee, line 4: Ditto columns ?

src/components/Security/Roles/Page.vue
berthieresteban: We are free to set cols<=12 and its very usefull when we set a property like `justify="center"` to center horizontally the col in the row, wich isn't possible if the col=12.
Also, if there is no other col in the same row, set cols<12 won't impact other cols in other rows.

PR #643, opened 2020-04-22
[KZL-1439] Create or update role

xbill82:
Migrates the Create or Update Role to Bootstrap v4

Create and update a role


berthieresteban:
Not sure if it's a bug.
If we want to remove an autorized action in a role: 
when updating if we delete the action it still in the role at true.
We must set that action to false.
We could check the new payload and force to false the deleted actions
Or Maybe we should just display that information somewhere...

xbill82:
@berthieresteban it's a bug, we should replace the role instead of just updating it

PR #657, opened 2020-04-27
reset password form

benoitvidis:
## What does this PR do ?

This PR implements the new up-coming features from https://github.com/kuzzleio/kuzzle-plugin-auth-passport-local/pull/70.

- implement a reset password form
- handle `expired_password` and `must_change_password` errors on login and redirect to reset password form in this case
- make sure other custom errors are properly handled

```
git clone -b password-policies git@github.com:kuzzleio/kuzzle-plugin-auth-passport-local.git
docker-compose up -d
docker-compose exec kuzzle sh -c 'echo \'{
  "plugins": {
    "kuzzle-plugin-auth-passport-local": {
      "passwordPolicies": [
        {
          "appliesTo": "*",
          "expiresAfter": "42 mn",
          "forbidLoginInPassword": true,
          "forbidReusedPasswordCount": 24,
          "mustChangePasswordIfSetByAdmin": true,
          "passwordRegex": ".{5,}"
        }
      ]
    }
  }
}\' > /var/app/.kuzzlerc && touch /var/app/plugins/available/kuzzle-plugin-auth-passport-local/index.js'
```

To get a reset token: 
```
curl -s http://localhost:7512/_plugin/kuzzle-plugin-auth-passport-local/password/resetToken/admin | docker run -i stedolan/jq -r '.result.resetToken'
```

You can now access the reset password form on `/#/reset-password/<token>`

When `expired_password` or `must_change_password` error received on login:

image

Direct access (for instance when coming from a "forgot password" email):

image

image

image

### DIFF
--- a/src/components/Common/Login/ResetPasswordForm.vue
+++ b/src/components/Common/Login/ResetPasswordForm.vue
@@ -1,0 +1,101 @@
+<template>
+  <form id="resetPasswordForm" method="post" @submit.prevent="resetPassword()">
+    <div class="resetPasswordForm-inputs">
+      <b-form-group
+        label="New password"
+        label-for="password"
+        label-cols-sm="4"
+        label-cols-md="3"
+      >
+        <b-form-input
+          autofocus
+          class="validate"
+          data-cy="ResetPassword-password"
+          id="password"
+          name="password"
+          required
+          tabindex="1"
+          type="password"
+          v-model="password"
+        />
+      </b-form-group>
+
+      <b-form-group
+        description="Re-type the password for confirmation"
+        label="Confirm password"
+        label-cols-sm="4"
+        label-cols-lg="3"
+        label-for="password2"
+      >
+        <b-input
+          class="validate"
+          data-cy="ResetPassword-password2"
+          id="password2"
+          name="password2"
+          required
+          tabindex="2"
+          type="password"
+          v-model="password2"
+        />
+      </b-form-group>
+
+      <div class="ResetPasswordForm-error" v-if="error">
+        <b-alert variant="danger" show dismissible>
+          Error: {{ error }}
+        </b-alert>
+      </div>
+
+      <div class="ResetPasswordForm-buttons float-right mt-3">
+        <b-button
+          variant="primary"
+          data-cy="ResetPassword-submitBtn"
+          type="submit"
+          name="action"
+          tabindex="3"
+          >Send</b-button
+        >
+      </div>
+    </div>
+  </form>
+</template>
+
+<script>
+export default {
+  name: 'ResetPasswordForm',
+  props: {
+    resetToken: String
+  },
+  data() {
+    return {
+      error: '',
+      password: '',
+      password2: ''
+    }
+  },
+  methods: {
+    async resetPassword() {
+      this.error = ''
+
+      if (this.password.trim() === '' || this.password2.trim() === '') {
+        this.error = 'All fields are mandatory'
+        return
+      }
+
+      if (this.password !== this.password2) {
+        this.error = 'Passwords do not match'
+        return
+      }
+
+      try {
+        await this.$store.direct.dispatch.auth.doResetPassword({
+          password: this.password,
+          token: this.$route.params.token
+        })
+        this.$emit('reset-password::after')
+      } catch (error) {
+        this.error = error.message
+      }
+    }
+  }
+}
+</script>
--- a/src/components/Common/Login/ResetPasswordForm.vue
+++ b/src/components/Common/Login/ResetPasswordForm.vue
@@ -1,7 +1,12 @@
 <template>
-  <form id="resetPasswordForm" method="post" @submit.prevent="resetPassword()">
+  <form
+    id="resetPasswordForm"
+    method="post"
+    @submit.prevent="resetPassword()"
+  >
     <div class="resetPasswordForm-inputs">
       <b-form-group
+        invalid-feedback="Password must not be empty"
         label="New password"
         label-for="password"
         label-cols-sm="4"
@@ -13,10 +18,12 @@
           data-cy="ResetPassword-password"
           id="password"
           name="password"
+          pattern=".*[^ ].*"
           required
           tabindex="1"
           type="password"
           v-model="password"
+          @blur="setValidated"
         />
       </b-form-group>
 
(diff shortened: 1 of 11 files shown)


src/components/Common/Login/ResetPasswordForm.vue
```diff
@@ -0,0 +1,101 @@
...
+      error: '',
+      password: '',
+      password2: ''
+    }
+  },
+  methods: {
+    async resetPassword() {
+      this.error = ''
+
+      if (this.password.trim() === '' || this.password2.trim() === '') {
```
xbill82, line 79: You may want to use native [BootstrapVue form validation](https://bootstrap-vue.org/docs/components/form-group#validation-state-feedback) here

src/components/Common/Login/ResetPasswordForm.vue
```diff
@@ -0,0 +1,101 @@
...
+
+      if (this.password !== this.password2) {
+        this.error = 'Passwords do not match'
+        return
+      }
+
+      try {
+        await this.$store.direct.dispatch.auth.doResetPassword({
+          password: this.password,
+          token: this.$route.params.token
```
xbill82, line 92: It is always better to decouple components from the routes [by making the router pass parameters as props](https://router.vuejs.org/guide/essentials/passing-props.html)

src/components/ConnectionAwareContainer.vue
```diff
@@ -107,6 +107,10 @@ export default {
       }
     },
     async authenticationGuard() {
+      if (this.$route.meta.skipLogin) {
```
xbill82, line 110: Ditto (on props from router)

src/components/ResetPassword.vue
```diff
@@ -0,0 +1,55 @@
...
+                alt="Welcome to the Kuzzle Admin Console"
+                class="mb-5 img-fluid"
+              />
+            </div>
+
+            <b-alert
+              class="text-center"
+              variant="warning"
+              data-cy="resetPasswordAlert"
+              :show="$route.params.showIntro"
```
xbill82, line 19: :point_up: props from router

src/components/ResetPassword.vue
```diff
@@ -0,0 +1,55 @@
...
+            >
+              <b>Warning!</b> You must update your password to continue
+            </b-alert>
+          </b-card-body>
+
+          <reset-password-form
+            :reset-token="$route.params.token"
```
xbill82, line 26: :point_up: props from router

src/components/Common/Login/ResetPasswordForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        })
+        this.$emit('reset-password::after')
+      } catch (error) {
+        this.error = error.message
+      }
```
berthieresteban, line 97: We probably should not display the kuzzle error returned since it might be not enough human readable.
To be consistent with the others catch blocks of the AC we should log that error in the console and display a toast
```suggestion
        this.error = 'An error occured.'
        console.error(error.message);
        this.$bvToast.toast(
            'The complete error has been printed to the console.',
            {
              title: err.message,
              variant: 'danger',
              toaster: 'b-toaster-bottom-right',
              appendToast: true
            }
          )
      }
```

src/components/ResetPassword.vue
```diff
@@ -0,0 +1,55 @@
...
+  },
+  methods: {
+    onReset() {
+      if (this.$store.getters.routeBeforeRedirect) {
+        this.$router.push({ name: this.$store.getters.routeBeforeRedirect })
+      } else {
+        this.$router.push('/').catch(() => {})
+      }
+
+      this.$store.direct.commit.routing.setRouteBeforeRedirect(undefined)
```
berthieresteban, line 51: + We can't be sure that the commit is executed after that if/else because we can't predict when the component will be destroy after the router push (ping @xbill82 ?)
+ Unecessary empty catch ? 
```suggestion
      const routeBeforeRedirect = this.$store.getters.routeBeforeRedirect;
      if (routeBeforeRedirect) {
        this.$store.direct.commit.routing.setRouteBeforeRedirect(undefined)
        this.$router.push({ name: routeBeforeRedirect })
      } else {
        this.$router.push('/')
      }
```

src/components/Common/Login/ResetPasswordForm.vue
xbill82: There are some cases where the Kuzzle error is just fine. Ben showed some screenshots where the errors where readable enough

benoitvidis: I only cought this behaviour on the signup component, but I believe it should be changed everywhere.

Error messages (not only Kuzzle ones) are supposed to be human-readable. 

I can only see situations where the users should get it:

- Either the message is useful to him, like a validation error, to fix the situation
- Or the error cannot be fixed by the user (bug, configuration error, third party down etc.), in which case the error message should be reported to the support.

Asking the user to open the development console to get it is probably a blocker for some users.

xbill82: Yes, the best practice should be:
* intercept the error and parse the error code
* display a toast with a human-readable message
* report the original Kuzzle error to the console, so that the user can report it to the support team if necessary

src/components/ResetPassword.vue
benoitvidis: I will update the code as it is confusing but the commit will always be fired with the current version.
The Javascript event loop won't jump to a next section within a synchronous function.

xbill82: good point

src/components/ConnectionAwareContainer.vue
benoitvidis: I am sorry, I did not manage to turn this one into a prop.

The problem is this component is loaded by the router and the business component is not aware of it and thus, cannot pass data to it.

src/components/Common/Login/ResetPasswordForm.vue
```diff
@@ -0,0 +1,114 @@
...
+          password: this.password,
+          token: this.resetToken
+        })
+        this.$emit('reset-password::after')
+      } catch (error) {
+        this.error = error.message
+      }
+    },
+    setValidated(event) {
+      event.target.parentNode.classList.add('was-validated')
```
xbill82, line 110: Oh... I find this a little bit weird... Why not use the `:state` dynamic attribute of `<b-input>`?

PR #670, opened 2020-04-28
[KZL-1512] Allow see/update profile properties than just the policies

Njuelle:
## What does this PR do ?
Allow see/update profile properties than just the policies

### Other changes
* Update kuzzle SDK version
* Add Cypress in dev dependencies and upgrade version
* Fix failings tests

### Screenshots
image


tests/e2e/cypress/integration/docs.spec.js
```diff
@@ -250,7 +250,7 @@ describe('Document List', function() {
     cy.get('.col > .col > .col > .Autocomplete > input').click()
     cy.get(
       '.col > .col > .Autocomplete > .Autocomplete-results > .Autocomplete-result'
-    ).click()
```
xbill82, line 253: Wouldn’t it be better to use the nth-child selector like you did in the next file?

PR #681, opened 2020-05-04
Change document fields order

### LINKED ISSUE #679: Displays the _kuzzle_info field in last position
Actually the `_kuzzle_info` field can be anywhere inside the document body.  

Since it represents Kuzzle internal metadata, it may be better to display this property in last position so user can see their document content first and then Kuzzle metadata.

We still want to keep them in the document body so user know that they can make query on it.

Also, the `id` should not be present inside the document body since a document can contain a `id` or even a `_id` field.

image

On this example, the displayed JSON should be
```
{ 
  age: 48
  name: "Test-48",
  _ kuzzle_info: { ... }
}
```

Aschen:
## What does this PR do ?

Fix #679 

Display `_kuzzle_info` field at the end and remove the `id` field.

Also format dates from `_kuzzle_info` like other date fields.

**Before**
image

**After**
image

## Other changes

 - add missing notification types

### DIFF
--- a/src/components/Data/Documents/DocumentListItem.vue
+++ b/src/components/Data/Documents/DocumentListItem.vue
@@ -67,7 +67,7 @@
         v-model="expanded"
         class="ml-3 DocumentListItem-content w-100"
       >
-        <pre v-json-formatter="{ content: formatedDocument, open: true }" />
+        <pre v-json-formatter="{ content: formattedDocument, open: true }" />
       </b-collapse>
     </b-row>
   </b-container>
--- a/src/components/Data/Documents/DocumentListItem.vue
+++ b/src/components/Data/Documents/DocumentListItem.vue
@@ -67,13 +67,14 @@
         v-model="expanded"
         class="ml-3 DocumentListItem-content w-100"
       >
-        <pre v-json-formatter="{ content: document, open: true }" />
+        <pre v-json-formatter="{ content: formatedDocument, open: true }" />
       </b-collapse>
     </b-row>
   </b-container>
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


src/components/Data/Documents/DocumentListItem.vue
```diff
@@ -67,13 +67,14 @@
         v-model="expanded"
         class="ml-3 DocumentListItem-content w-100"
       >
-        <pre v-json-formatter="{ content: document, open: true }" />
+        <pre v-json-formatter="{ content: formatedDocument, open: true }" />
```
xbill82, line 70: :gb:
```suggestion
        <pre v-json-formatter="{ content: formattedDocument, open: true }" />
```

src/components/Data/Documents/DocumentListItem.vue
```diff
@@ -121,6 +122,11 @@ export default {
     },
     checkboxId() {
       return `checkbox-${this.document.id}`
+    },
+    formatedDocument () {
```
xbill82, line 126: :gb:
```suggestion
    formattedDocument () {
```

xbill82: Also, it would be probably nice to add a comment to explain the motivation of this transform, which might seem weird to a developer that sees this in a year ("da hell is that? KLEAN KODEZ ATTITUDE LOL! Let's delete this")

xbill82: (could perfectly be me, btw)

xbill82:
Please, @github add the "Merge after CI succeeds" button

PR #686, opened 2020-05-05
Fix advanced search

### LINKED ISSUE #666: [AdvancedSearch] Generate query based on field datatype
## Expected Behavior
```gherkin
Given a field `dummy` of type `keyword` in the collection mapping
When I select the `dummy` field in the Advanced Query panel
And I select the "Match" operator
And I type a string in the value field
And I press the "Search" button
Then I expect the search to display the given results
```
## Current Behavior
I see an error toast and the following error in the browser console
```
Error: "failed to create query: {
  "bool" : {
    "should" : [
      {
        "bool" : {
          "must" : [
            {
              "match_phrase_prefix" : {
                "dummy" : {
                  "query" : "toto",
                  "slop" : 0,
                  "max_expansions" : 50,
                  "boost" : 1.0
                }
              }
            }
          ],
          "adjust_pure_negative" : true,
          "boost" : 1.0
        }
      }
    ],
    "adjust_pure_negative" : true,
    "boost" : 1.0
  }
}"
```

## Possible Solution
The Advanced Search might generate `match_phrase_prefix` queries for `text` fields and `match` queries for `keywords`

## Context (Environment)

Kuzzle version: v2.x
Admin Console Version: v3.x, v4.x

xbill82:
## What does this PR do ?
Fixes the advanced search in a pretty hacky way.

* create a collection with a `text` and a `keyword` field
* perform searches against each field using the Advanced search
* check you always get results

### Boyscout
Added tests for previous hotfix (on collecion creation)


tests/e2e/cypress/integration/search.spec.js
```diff
@@ -66,7 +53,7 @@ describe('Search', function() {
     )
   })
 
-  it('perists the Quick Search query in the URL', function() {
+  it('perists the Quick Search query in the URL', function () {
```
Leodau, line 56:
```suggestion
  it('persists the Quick Search query in the URL', function () {
```

Issue #659, opened 2020-04-28
Does not hide filters when searching for users or profiles


Aschen:
When I select a role, the whole UI disappear and I see the spinner. Only the result list should be replaced by the spinner and not the filters.

When I search an user, the whole UI disappear and I see the spinner. Only the result list should be replaced by the spinner and not the filters.

PR #693, opened 2020-05-07
Add keyboard navigation through indexes and collections

Aschen:
## What does this PR do ?

This PR allows users to navigate through indexes and collections and then documents with keyboard inputs only.  

The index, collection filters and document QuickFilter are now instances of `AutoFocusInput` which listen to keyboard event to directly focus the corresponding input.  
Then this component can emit a `submit` event when `Enter` is pressed.

The index, collection and document list pages listen for this submit event to navigate to the first item of the filtered list.

navigate-keyboard

### Other changes

 - `DocumentListItem` is now responsible of pushing the `Update` route to the router when the update button is clicked
 - Fix collection breadcrumb
 - Fix https://github.com/kuzzleio/kuzzle-admin-console/issues/659

### DIFF
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -19,6 +19,10 @@ export default {
       type: String,
       default: ''
     },
+    activated: {
+      type: Boolean,
+      default: true
+    }
   },
   computed: {
     refName() {
@@ -72,9 +76,17 @@ export default {
     }
   },
   mounted () {
+    if (! this.activated) {
+      return
+    }
+
     this.listenKeypress()
   },
   beforeDestroy() {
+    if (! this.activated) {
+      return
+    }
+
     this.stopListenKeypress()
   }
 
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -34,11 +34,9 @@ export default {
   },
   methods: {
     looseFocus() {
-      console.log('loose foc')
       this.isFocus = false
     },
     listenKeypress () {
-      console.log('listen')
       document.onkeypress = keyEvent => {
         if (this.isFocus) {
           if (keyEvent.code === 'Enter') {
@@ -54,7 +52,6 @@ export default {
       }
     },
     stopListenKeypress () {
-      console.log('stop')
       document.onkeypress = null
     }
   },
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -91,7 +91,6 @@ export default {
       document.onkeypress = null
     }
   },
-  //Cannot mount AutoFocusInput at: KuzzleAdminConsole > ConnectionAwareContainer > Home > DataLayout > DocumentsPage > Filters > QuickFilter > AutoFocusInput. An AutoFocusInput is already mounted at: KuzzleAdminConsole > ConnectionAwareContainer > Home > DataLayout > DocumentsPage > AutoFocusInput.
   mounted() {
     if (componentParents.length !== 0) {
       throw new Error(`Cannot mount AutoFocusInput at: ${getParents(this).join(' > ')}. \nAn AutoFocusInput is already mounted at: ${componentParents.join(' > ')}.`)
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -15,6 +15,10 @@ export default {
       type: String,
       required: true
     },
+    initialValue: {
+      type: String,
+      default: ''
+    },
   },
   computed: {
     refName() {
@@ -28,7 +32,7 @@ export default {
   },
   data () {
     return {
-      value: '',
+      value: this.initialValue,
       isFocus: false
     }
   },
@@ -49,6 +53,18 @@ export default {
         this.isFocus = true;
         this.value = ''
         this.$refs[this.refName].focus()
+
+        // Firefox does not follow the key event to the input but Chrome does
+        // so after the event has finished his propagation, we check the input
+        // value and on Firefox it will be empty so we put the first letter inside
+        // Using preventDefault() to manually input the first letter and prevent
+        // to have double input on Chrome doesn't work well because then the cursor
+        // is at the wrong place
+        setTimeout(() => {
+          if (this.value === '') {
+            this.value = keyEvent.key
+          }
+        }, 1)
       }
     },
     stopListenKeypress () {
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -3,8 +3,7 @@
     :ref="refName"
     v-model="value"
     v-bind="$attrs"
-    @blur="looseFocus"
-    data-auto-focus="true"
+    data-auto-focus-input="true"
   />
 </template>
 
@@ -24,7 +23,6 @@ export default {
   data() {
     return {
       value: this.initialValue,
-      isFocus: false
     }
   },
   computed: {
@@ -33,27 +31,27 @@ export default {
     }
   },
   methods: {
-    // AutoFocusInput is only activated when user is not using another input
-    shouldAcquireFocus() {
+    // AutoFocusInput should not handle keypress when user is using another input
+    shouldHandleKeypress() {
       const tagType = document.activeElement.tagName
 
       // If we are currently selecting another input
-      if (tagType === 'INPUT' && !document.activeElement.dataset.autoFocus) {
+      if (tagType === 'INPUT' && !document.activeElement.dataset.autoFocusInput) {
         return false
       }
 
       return true
     },
-    looseFocus() {
-      this.isFocus = false
+    isFocus() {
+      return this.$refs[this.refName].$refs.input === document.activeElement
     },
     listenKeypress() {
       document.onkeypress = keyEvent => {
-        if (!this.shouldAcquireFocus()) {
+        if (!this.shouldHandleKeypress()) {
           return
         }
 
-        if (this.isFocus) {
+        if (this.isFocus()) {
           if (keyEvent.code === 'Enter') {
             this.$emit('submit')
           }
@@ -61,7 +59,6 @@ export default {
           return
         }
 
-        this.isFocus = true
         this.$refs[this.refName].focus()
 
         // Firefox does not follow the key event to the input but Chrome does
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -4,6 +4,7 @@
     v-model="value"
     v-bind="$attrs"
     @blur="looseFocus"
+    data-auto-focus="true"
   />
 </template>
 
@@ -18,10 +19,6 @@ export default {
     initialValue: {
       type: String,
       default: ''
-    },
-    activated: {
-      type: Boolean,
-      default: true
     }
   },
   computed: {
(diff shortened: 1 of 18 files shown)


src/components/Common/AutoFocusInput.vue
```diff
@@ -0,0 +1,97 @@
...
+      if (tagType === 'INPUT' && !document.activeElement.dataset.autoFocus) {
+        return false
+      }
+
+      return true
+    },
+    looseFocus() {
+      this.isFocus = false
+    },
+    listenKeypress () {
```
Njuelle, line 55: you can directly listen for keys events with a `v-on` directive on components
https://vuejs.org/v2/guide/events.html#Key-Modifiers

Aschen: Yes but I want to catch every event so that why the handler is on the `document` object.

src/components/Common/AutoFocusInput.vue
```diff
@@ -0,0 +1,97 @@
...
+    name: {
+      type: String,
+      required: true
+    },
+    initialValue: {
+      type: String,
+      default: ''
+    }
+  },
+  computed: {
```
xbill82, line 24: NITP: Usually, we follow a convention for the order of the members of a Vue Single File Component:
* name
* components
* directives
* props
* data
* computed
* methods
* mounted
* watch
* (other lifecycle hooks)

src/components/Common/AutoFocusInput.vue
```diff
@@ -0,0 +1,97 @@
...
+  },
+  data () {
+    return {
+      value: this.initialValue,
+      isFocus: false
+    }
+  },
+  methods: {
+    // AutoFocusInput is only activated
+    enable () {
```
xbill82, line 42: I don't understand this method: looks like it should be a getter that tells whether the element `isEnabled`, but then it should be named `isFocused`. But then, in this case, I'd just go like

```javascript
return this.refs[this.refName].hasFocus()
```

src/components/Common/AutoFocusInput.vue
```diff
@@ -0,0 +1,97 @@
...
+      const tagType = document.activeElement.tagName
+
+      // If we are currently selecting another input
+      if (tagType === 'INPUT' && !document.activeElement.dataset.autoFocus) {
+        return false
+      }
+
+      return true
+    },
+    looseFocus() {
```
xbill82, line 52: Same here: should be a getter, I think.

src/components/Common/AutoFocusInput.vue
```diff
@@ -0,0 +1,97 @@
...
+        this.isFocus = true;
+        this.$refs[this.refName].focus()
+
+        // Firefox does not follow the key event to the input but Chrome does
+        // so after the event has finished his propagation, we check the input
+        // value and on Firefox it will be empty so we put the first letter inside
+        // Using preventDefault() to manually input the first letter and prevent
+        // to have double input on Chrome doesn't work well because then the cursor
+        // is at the wrong place
+        setTimeout(() => {
```
xbill82, line 78: This whole thing looks very convoluted to me, or I don't understand what we're trying to do.

src/components/Common/Filters/QuickFilter.vue
```diff
@@ -109,35 +109,37 @@ export default {
     submitOnType: {
       type: Boolean,
       default: true
+    },
+    initialValue: {
```
xbill82, line 113: Thank you for renaming all this crap.

src/components/Common/Filters/QuickFilter.vue
```diff
@@ -109,35 +109,37 @@ export default {
...
   data() {
     return {
-      inputSearchTerm: this.searchTerm
+      value: this.initialValue
     }
   },
   methods: {
     submitSearch() {
-      this.$emit('update-filter', this.inputSearchTerm)
+      this.$emit('update-filter', this.value)
```
xbill82, line 125: By the way this event should be renamed to `filter-submitted`

src/components/Common/Filters/QuickFilter.vue
```diff
@@ -109,35 +109,37 @@ export default {
...
-        return
-      }
-      this.submitSearch()
-    },
-    searchTerm: {
-      immediate: true,
-      handler(val) {
-        this.inputSearchTerm = val
+    value () {
+      this.$emit('input', this.value)
```
xbill82, line 139: And YES! We should maximize the use of the `input` event and the `value` prop, so that components using our custom components can leverage the `v-model` macro-attribute

src/components/Data/Documents/DocumentListItem.vue
```diff
@@ -147,7 +147,10 @@ export default {
     },
     editDocument() {
       if (this.canEdit) {
-        this.$emit('edit', this.document.id)
+        this.$router.push({
```
xbill82, line 150: I don't agree with this. Such a "leaf" component shouldn't take the responsibility of pushing the whole app to a new route (thus be coupled to the route naming). Parent component should be in charge of deciding what happens when the user asks to edit a document.

src/components/Data/Documents/Page.vue
```diff
@@ -442,6 +432,18 @@ export default {
       this.fetchDocuments()
     },
     performSearchDocuments,
+    navigateToDocument () {
```
xbill82, line 435: And this is good! The big `Page` component is the one who handles the router.

src/components/Materialize/Modal.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <div v-if="isOpen" :id="id" class="modal" :class="computedClasses">
+    <div v-if="isOpen" :id="id" class="modal" :class="computedClasses" data-modal="true">
```
xbill82, line 3: This component should die.

xbill82: Like ALL the components in the `Materialize` directory.

xbill82:
... and ...
End-to-end tests need to be written for the implemented features. :man_shrugging:

src/components/Common/AutoFocusInput.vue
Aschen: This method is to prevent to focus the AutoFocusInput if the user is focusing another input.  

I cannot use a computed because I need to get the current active element from the DOM each time and computed will cache the result and never compute it again since I don't change the state.  

I renamed it to `shouldAcquireFocus` for more clarity

src/components/Common/AutoFocusInput.vue
Aschen: I didn't understand because this method only set the `isFocus` property?

src/components/Common/AutoFocusInput.vue
Aschen: As the comment say, I have a different behavior on Chrome and Firefox and I use this tricks to make it works in both.

An AutoFocusInput is an input which will be auto focused in the user start to type on his keyboard.  
For example on Github if you start to type without focusing any input then you will autofocus the Search Bar

src/components/Data/Documents/DocumentListItem.vue
Aschen: Why it shouldn't?  I thought the component approach was made for this, to have more "independent" components.  

Here I don't see why the parent should be in charge of this. Everywhere this component is rendered it has the same meaning and if the user click on update then we should redirect him to the update page.

Otherwise we have to pass method and event alongside the component chain which is a very tedious task.

IMHO the parent component should not have the responsibility of the navigation either, like applying some rights or idk. This should only be in the router, what do you think?

src/components/Materialize/Modal.vue
Aschen: Why they are not deleted so?

src/components/Common/AutoFocusInput.vue
xbill82: I mean
```
computed: {
  isFocus() {
    return this.refs[this.refName].isFocus
  }
}
```
And let the DOM manage the focus state. Or I'm probably missing something fundamental about the purpose of this component.

src/components/Common/AutoFocusInput.vue
xbill82: Ok, now I get the purpose of this whole component. Still... I'd say the responsibility of auto-focusing an input should be of the root component on the page, rather than a single component. Like, what if someone adds many `AutoFocusInput`s in the same page?

src/components/Common/AutoFocusInput.vue
Aschen: I had to use a method because computed property are only updated on state change

src/components/Common/AutoFocusInput.vue
Aschen: They shouldn't do that :sweat_smile:

Aschen: Eventually I can add a guard to prevent developer to make this mistake

xbill82: Well, I think this is not good. A leaf component shouldn't have the responsibility of a scope greater than its own one. I'm still convinced that the auto-focus is a concern of the page, not the input element. To me, this component should not exist and the auto-focus logic should be moved to the page.

src/components/Data/Documents/DocumentListItem.vue
xbill82: Exactly, components should be independent and the idea is to minimize coupling, especially with the route names and params, because they're totally application-specific.
Leaf components should be kept as light as possible and IMHO the good practice is always to ask yourself if the same component would work in another application.
The argument that we're not using this component in another application is not a valid one to justify coupling.

Root-ish components are usually more coupled to the application-specific logic and it is more acceptable to make them aware of routes, for example, especially because, most of the time, we want to tend to displace this logic toward the roots, so that we can play more easily with leaf component and keep the UI lean.

I'll give you an example.
Imagine that tomorrow we decide to move the document edition to a modal or an inline editor right in the same page instead of doing it in another page. It would be something the Page component would be aware, of course, and if we keep the leaf component away from this concern, we'll have to modify only the parent component.

xbill82: I agree that passing the method alongside the component tree is tedious, this is why I always try to keep things simple by making leaf components just `$emit` and the parent just listen to `@event="$router.push()`

I agree that, when it comes to situations where we have several degrees of component nesting, it is tedious to forward the events from children to parents. I think the good compromise is "try not to couple leaf components while trying not to have too much depth in the component tree".

src/components/Common/AutoFocusInput.vue
xbill82: ah good point

src/components/Materialize/Modal.vue
xbill82: Bc I forgot

src/components/Data/Documents/DocumentListItem.vue
xbill82: But I see your point: you are disturbed by having to pass such a tiny event from the Item :point_right: List :point_right: Page and I agree it's overkill to pass an item through 3 components just to go to another page. So, in this case, I think yours is the best solution.

src/components/Materialize/Modal.vue
Aschen: Those components are still used everywhere in the AC. IMHO this is outside the scope of this PR.

Aschen:
I added a check to ensure this component is never mount two times on the same page

image

test/e2e/cypress/integration/indexes.spec.js
```diff
@@ -64,4 +64,15 @@ describe('Indexes', () => {
...
+
+  it('Should be able to autofocus index search', () => {
+    cy.request('POST', 'http://localhost:7512/nyc-open-data/_create')
+    cy.request('POST', 'http://localhost:7512/iot-data/_create')
+
+    cy.waitOverlay()
+
+    cy.get('body').type('n{enter}')
+
+    cy.contains('This index has no collections.')
```
xbill82, line 76: Maybe you can add
```suggestion
    cy.contains('nyc-open-data')
    cy.contains('This index has no collections.')
```

To be sure you are looking at the right index

test/e2e/cypress/integration/collections.spec.js
```diff
@@ -102,4 +102,18 @@ describe('Collection management', function() {
...
+    cy.request('PUT', `${kuzzleUrl}/${indexName}/foobar`)
+
+    cy.waitOverlay()
+
+    cy.visit(`/#/data/${indexName}/`)
+    cy.wait(500)
+
+    cy.get('body').type('f{enter}')
+
+    cy.contains('This collection is empty')
```
xbill82, line 117: Same remark as the following test

PR #696, opened 2020-05-11
Fix lost of search results after a connection loss

morgandruesne:
## What does this PR do ?

Fix a bug where when a reconnection, the results of a search query are not available automatically.


src/components/Data/Documents/Page.vue
```diff
@@ -411,7 +413,10 @@ export default {
           this.index,
           this.collection
         )
-        this.fetchDocuments()
+        await waitForConnected(1000)
+        if (Vue.prototype.$kuzzle.protocol.state === 'connected') {
```
xbill82, line 417: You don't need to get `$kuzzle` from `Vue.prototype` as all the components are instances of the `Vue` class. You just need to do
```suggestion
        if (this.$kuzzle.protocol.state === 'connected') {
```

Issue #682, opened 2020-05-04
[POC] Simultaneous compatibility with multiple Kuzzle versions


xbill82:
We understand the interest of providing to our users an Admin Console that would be able to seamlessly connect to any supported Kuzzle backend, given its version is currently supported.

This is going to be a Proof-of-Concept, meant to let us understand the feasibility of this feature, namely
* importing multiple versions of the Kuzzle Javascript SDK 
* wrapping the SDK APIs with different upper-level classes
* instantiate the correct wrapper given the version of the backend
* model the behavior of the UI on the available features of the backend

Feel free to use this issue as a place for debating.

PR #697, opened 2020-05-12
Kuzzle multiversion proof-of-concept

xbill82:
## What does this PR do ?

Implements the proof-of-concept showing that the same instance of the Admin Console can connect to a Kuzzle v1.x and a Kuzzle v2.x (issue #682)

Be sure to have 2 instances of Kuzzle, one v1 and one v2. Create two different connections, one for each backend and specify the correct version for each backend. Verify you can connect, login and see the list of the indexes

### Important note
This is still in the POC phase, this the implementation is lacking many features. Don't expect the Admin Console to work beyond the scope of the POC, i.e. connecting, logging-in and listing the indexes.
This (very succint) work is meant to help us decide whether we take this path or not.

### TODO
- [x] Tests v2
- ~Tests v1~
- [x] Show consistent feedback when connecting to a Kuzzle of the wrong version
- ~CI with stages for v1 and v2~

### Boyscout
- Got rid of `vuex/common/crudlDocument`


src/services/kuzzleWrapper-v2.ts
```diff
@@ -0,0 +1,284 @@
...
+      try {
+        const res = await kuzzle.security.getCredentials(strategy, d._id)
+        u.credentials[strategy] = res
+      } catch (e) {
+        u.credentials[strategy] = {}
+      }
+    }
+
+    users.push(u)
+  }
```
Leodau, line 161: Is the[ id-length](https://eslint.org/docs/rules/id-length) of this project at 1? (d, u) Kinda cryptic.

xbill82: ROTFL! Ur right dude

src/vuex/modules/kuzzle/getters.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  },
+  currentEnvironment(state, getters) {
+    if (!getters.hasEnvironment) {
+      return null
+    }
+
+    if (!state.currentId) {
+      return null
+      // const firstKey = Object.keys(state.environments)[0]
```
Njuelle, line 40:
```suggestion
```

src/vuex/modules/kuzzle/getters.ts
```diff
@@ -0,0 +1,53 @@
...
+      // return state.environments[firstKey]
```
Njuelle, line 41:
```suggestion
```

src/vuex/modules/kuzzle/getters.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return state.environments[state.currentId]
+  },
+  hasEnvironment(state) {
+    return Object.keys(state.environments).length !== 0
+  },
+  environments(state) {
```
Njuelle, line 49: lol!

src/components/Common/Environments/CreateEnvironment.vue
```diff
@@ -130,13 +134,22 @@ export default {
...
-        ssl: useHttps
+        ssl: useHttps,
+        backendMajorVersion: 2
       },
-
       submitting: false
     }
   },
   computed: {
+    majorVersions() {
```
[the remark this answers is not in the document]
xbill82: Because `data` is for things that are meant to be mutable. I use computed values as constants.

src/components/Data/Documents/Page.vue
```diff
@@ -640,6 +634,10 @@ export default {
       this.$router.push({ query: mergedQuery }).catch(() => {})
     },
     addHumanReadableDateFields() {
+      if (!this.collectionMapping) {
```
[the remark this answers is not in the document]
xbill82: I ran into a case where the `collectionMapping` was `null` after the component tried to fetch it, which broke this function.

src/vuex/modules/kuzzle/getters.ts
xbill82: yep

src/components/Common/Environments/CreateEnvironment.vue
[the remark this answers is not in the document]
Njuelle: Isn't that a bit overkill?
why not just use a constant?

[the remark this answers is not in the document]
xbill82: moved to `data()`

berthieresteban:
With a kuzzle V1 running and an environment specifying kuzzle v2
On this screen, changing the environment version from 2 to 1 doesn't refresh the connection and we get stuck here (since there is no button to do something we need to refresh the page)
image

berthieresteban:
Creating an environment with the name of an existing environment won't display an error (it just quit like if the creation was a success) but doesn't create/update any environment (wich is a good point ^^)

src/components/Common/Environments/CreateEnvironment.vue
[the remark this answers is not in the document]
berthieresteban: just to feed the debate  (I have no idea of wich is the best/optimized practice for const data in a component)
image

src/components/Data/Documents/ViewMap.vue
```diff
@@ -38,10 +38,7 @@
           <dropdown :id="currentDocument.id" myclass="icon-black">
             <li>
               <a
-                v-title="{
-                  active: !canDelete,
-                  title: 'You are not allowed to delete this document'
-                }"
+                title="You are not allowed to delete this document"
                 :class="{ disabled: !canDelete }"
```
berthieresteban, line 42: I can't find where we described a `.disabled` class in the AC

test/e2e/cypress/integration/users.spec.js
```diff
@@ -1,5 +1,6 @@
 describe('Users', function() {
   const kuzzleUrl = 'http://localhost:7512'
+  const backendVersion = Cypress.env('BACKEND_VERSION') || 2
```
berthieresteban, line 3: Why is there a difference between here and the other `spec.js` files ? 
Also since the only use of `backendVersion` is with `initLocalEnv()` wich set a default value to that parameter it seems useless to do it another time here

PR #701, opened 2020-05-14
Refacto de la muerte (the return)

### LINKED ISSUE #688: Error loop when trying to connect to Kuzzle v1
## Expected Behavior
```gherkin
When I try to connect to a Kuzzle v1.x instance
And the instance is online
Then I get an error
And I see the connection selector
```

## Current Behavior
```gherkin
When I try to connect to a Kuzzle v1.x instance
And the instance is online
Then I see the spinner spinning
And the spinner goes round and round and round and round and round and round...
```

## Possible Solution
This is due to the connection retry loops in the `ConnectionAwareContainer.vue` and the `EnvironmentSelector.vue` components. If connection does not occur, they perform another attempt 2 seconds later no matter why the connection did not occur. We should check the error type and retry only if it's a network-related error

## Context (Environment)

Kuzzle version: 1.x
Admin Console Version: 4.x

xbill82:
## What does this PR do ?
Fixes #688 by introducing a consistent connection-error state in the `ConnectionAwareContainer`.

## Update
But this happens to be the Pandora's box and it evolved to the REFACTO DE LA MUERTE VERSIÓN DOS PUNTO CERO :skull_and_crossbones: .

### Details (if you're curious)
My initial aim was just to address the issue of having the AC retrying to open the connection to a Kuzzle that seemes unreachable and just giving up with a consistent feedback if the backend is online but an error occurs. Also, if the backend goes down during a session, the AC should display a toast that would disappear once the connection is recovered. This seemed trivial but it happened to be not (mostly because the previous implementation was bad).
Actually, the best way to implement this behaviour is to leverage the SDK events `connected`, `disconnected`, `reconnected` and `networkError`.
The tricky thing about connection errors is that they might or might not have an error code/id. If the error is triggered by the Kuzzle instance (e.g. the SDK version is incompatible), the error will have an ID, while if the error is triggered by the SDK (e.g. the network is unavailable) the error will _not_ have an ID.
Another tricky thing is that the `connect()` call to the SDK, doesn't really seem to trigger any error, even when the backend is unreachable.
Besides this and the refactoring of the `kuzzle` and `auth` store modules, the main difference with the previous implementation is the asynchronous fashion of the connection process. Now, the authentication is triggered by the `connected` event, while before it was synchronously triggered by the `switchEnvironment` action.

First, run
```sh
kourou instance:spawn -v 2
```
then
```sh
kourou instance:spawn -v 1
```

Then, launch the AC and import the following connections
```json
{
  "Local V2": {
    "name": "Local v2",
    "color": "darkblue",
    "host": "localhost",
    "port": 7512,
    "ssl": false,
  },
  "Unreachable Host": {
    "name": "Unreachable Host",
    "color": "magenta",
    "host": "aosdasd",
    "port": 7512,
    "ssl": false,
  },
  "Local V1": {
    "name": "Local v1",
    "color": "purple",
    "host": "localhost",
    "port": 7513,
    "ssl": false,
  },
  "Wrong V2": {
    "name": "Wrong V2",
    "color": "red",
    "host": "localhost",
    "port": 7513,
    "ssl": false,
  }
}
```

```gherkin
Given I have a Kuzzle v1 (port 7513) and a Kuzzle v2 (port 7512) running
When I connect to the "Local V2"
Then I get the normal connected state (Login or Home)
When I switch to the "Local V1"
Then I get the connection-error state telling me that the SDK version is not compatible
```

```gherkin
Given I have a Kuzzle v1 (port 7513) running
When I connect to the "Local V1"
Then I get the connection-error state telling me that the SDK version is not compatible
```

### Other changes
* Another thing I did (which I could have left for another PR, probably) is that the authentication state-machine now recognizes three states: not authenticated, authenticated as Anonymous and authenticated as a user. To make things more solid, all the `can*` utility functions have been transformed in a set of getters that automatically update their value when the auth-state changes.
* Again, concerning the auth, I restored the re-login modal upon token expiration. I also added a test for this.

### Boyscout
* [Adding a new connection doesn't mean switching to it](https://github.com/kuzzleio/kuzzle-admin-console/pull/701/files#diff-80dfd428ffd78afbb4adf34eff3909c5L66)
* Deleted some useless components

### To be continued
* We will need to add a set of cross-environment tests, with several instances of Kuzzle of different versions. This will need extensive study to setup a tough CI.


src/components/Error/KuzzleErrorPage.vue
```diff
@@ -1,37 +1,34 @@
...
+    <b-container>
+      <b-jumbotron>
+        <template v-slot:header>
+          <img
+            alt="Welcome to the Kuzzle Admin Console"
+            class="mb-3"
+            height="60"
+            src="../../assets/logo.svg"
+          />
+          <h2>Something went wrong connecting to Kuzzle</h2>
```
rolljee, line 12: >Something went wrong connecting to Kuzzle > `Something went wrong while connecting to Kuzzle`?

src/vuex/modules/common/kuzzle/store.ts
```diff
@@ -31,6 +31,13 @@ export const envColors = [
   'magenta'
 ]
 
+const wait = async ms =>
```
rolljee, line 34: Unnecessary async i guess

src/vuex/modules/common/kuzzle/store.ts
```diff
@@ -31,6 +31,13 @@ export const envColors = [
...
+  new Promise(resolve => {
+    setTimeout(() => {
+      resolve()
+    }, ms)
+  })
+
 const checkEnvironment = e => {
```
rolljee, line 41: missing async since their are await statement in this fonction

xbill82: Where do you see await statements in this function?

alexandrebouthinon: @rolljee in splited code review mode you need to unfold parts sometimes. The await statements you seen is part of 
```js
async switchEnvironment(context, id) {
```

src/components/ConnectionAwareContainer.vue
```diff
@@ -37,13 +45,15 @@
 </template>
 
 <script>
```
xbill82, line 47: This is the component where all the magic happens (LOL)

src/components/ConnectionAwareContainer.vue
```diff
@@ -93,48 +134,43 @@ export default {
...
-        if (await this.$store.direct.dispatch.auth.checkToken()) {
-          this.$log.debug('ConnectionAwareContainer::Token bueno')
-        } else {
-          this.$log.debug('ConnectionAwareContainer::Token no bueno')
-          this.$router.push({ name: 'Login', query: { to: this.$route.name } })
-        }
-      } catch (error) {
-        this.$log.debug('ConnectionAwareContainer::Token no bueno (error)')
-        this.$log.error(error)
+      this.$log.debug('ConnectionAwareContainer::authentication guard')
```
xbill82, line 151: TODO This should redirect only if the current page needs authentication

xbill82:
TODO test invalid token modal

src/components/Data/Documents/Page.vue
```diff
@@ -556,11 +556,6 @@ export default {
 
     // PERMISSIONS
     // =========================================================================
```
Leodau, line 558: Empty comment block?

src/components/Data/Collections/Dropdown.vue
```diff
@@ -122,15 +117,16 @@ export default {
     myclass: String
   },
   computed: {
-    // WARNING THIS IS EVIL
+    ...mapGetters('auth', [
+      'canEditCollection',
+      'canSubscribe',
+      'canTruncateCollection'
+    ]), // WARNING THIS IS EVIL
```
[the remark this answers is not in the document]
xbill82: haha you're SOOOOO right.
I don't remember why I wrote this. I guess it's legacy stuff that never got cleaned.

PR #705, opened 2020-05-15
Error subcodes

morgandruesne:
## What does this PR do ?
Use error subcode rather than regex-ing the error message in the `Home` page

### DIFF
--- a/src/components/Home.vue
+++ b/src/components/Home.vue
@@ -76,11 +76,9 @@ export default {
     noop() {}
   },
   mounted() {
-    const invalidTokenRegex = /Invalid token/
     this.$kuzzle.on('tokenExpired', () => this.onTokenExpired())
     this.$kuzzle.on('queryError', e => {
-      // TODO use the error codes when available
-      if (invalidTokenRegex.test(e.message)) {
+      if (e.id === 'security.token.invalid') {
         this.onTokenExpired()
         this.tokenExpiredIsOpen = true
       }


xbill82:
Ok, wait... I thought we could access the Kuzzle error code from the SDK error but I can't find any clue of it neither [here](https://docs.kuzzle.io/sdk/js/7/essentials/error-handling/) nor [here](https://github.com/kuzzleio/sdk-javascript/blob/master/src/KuzzleError.js).

Is `code` what we're looking for?

morgandruesne:
it's not here? https://docs.kuzzle.io/core/2/api/essentials/error-handling/

xbill82:
:astonished: didn't see it... Maybe my page wasn't up to date (or probably my brain needed a reboot...)

src/components/Home.vue
```diff
@@ -76,11 +76,9 @@ export default {
     noop() {}
   },
   mounted() {
-    const invalidTokenRegex = /Invalid token/
     this.$kuzzle.on('tokenExpired', () => this.onTokenExpired())
     this.$kuzzle.on('queryError', e => {
-      // TODO use the error codes when available
-      if (invalidTokenRegex.test(e.message)) {
+      if (e.code === 401) {
```
xbill82, line 81: Ok, my bad... I meant we should use the `id` attribute since it's more human-readable.

src/components/Home.vue
```diff
@@ -78,7 +78,7 @@ export default {
   mounted() {
     this.$kuzzle.on('tokenExpired', () => this.onTokenExpired())
     this.$kuzzle.on('queryError', e => {
-      if (e.code === 401) {
+      if (e.id === "0x07010001") {
```
xbill82, line 81: Using `security.token.invalid` doesn't work?

alexandrebouthinon: Yes @morgandruesne, hexadecimal formats are not intend to be used on "high level application" but only when error size (in bytes) is a prerequisite for low level applications (like embedded software maxed to 32 bits in average)

morgandruesne: I'm gonna do this ! thanks for the explanation :smile:

PR #707, opened 2020-05-18
[CI] fix travis.yml for AC v3

Njuelle:
## What does this PR do ?
in accordance with Travis documentation for deployment on an S3 bucket, added the missing "upload_dir" parameter and fixed the "/"  that  should not be present for this parameter

just review carefully changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ jobs:
 
       script:
         - npm install
-        - CYPRESS_RETRIES=5 npm run e2e
+        - CYPRESS_RETRIES=5 npm run test:e2e
 
     # ---------------------------------------
     # Regular Build & Deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ jobs:
         - npm run lint
 
     # ---------------------------------------
-    # Functionnal Testing
+    # End-to-end Testing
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         - npm run lint
 
     # ---------------------------------------
-    # Functionnal Testing
+    # End-to-end Testing
```
xbill82, line 44: THANKS

.travis.yml
```diff
@@ -59,15 +59,56 @@ jobs:
     # ---------------------------------------
     # Regular Build & Deploy
     # ---------------------------------------
-    # console.kuzzle.io
+    # console.kuzzle.io/kuzzle-v1
```
xbill82, line 62: Sorry, why don't we take this out of the v3 branches?

Njuelle: For now I just wanted to fix what was broken without touching the travis.yml too much.

xbill82: I understand... I'd be confident, since the PR points on the dev branch and, in the worst case, the damage is done on next-console, what do you think?

Njuelle: yes, I agree ! I will clean up this travis file ;)

PR #719, opened 2020-06-03
Use regex for contains search

morgandruesne:
## What does this PR do ?
This PR allows you to perform contains searches on string and keyword fields.
Changing match_phrase_prefix for regexp 

### Other changes
rename filter attribute match by contains

### DIFF
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -424,28 +424,28 @@ export default {
     // LIST (FETCH & SEARCH)
     // =========================================================================
     async loadAllTheThings() {
-      this.loading = true
       try {
+        this.loading = true
         await this.loadMappingInfo()
+        this.loadListView()
+        this.saveListView()
+
+        this.currentFilter = filterManager.load(
+          this.index,
+          this.collection,
+          this.$route
+        )
+        filterManager.save(
+          this.currentFilter,
+          this.$router,
+          this.index,
+          this.collection
+        )
+        await this.fetchDocuments()
+        this.loading = false
       } catch {
         this.loading = false
       }
-      this.loadListView()
-      this.saveListView()
-
-      this.currentFilter = filterManager.load(
-        this.index,
-        this.collection,
-        this.$route
-      )
-      filterManager.save(
-        this.currentFilter,
-        this.$router,
-        this.index,
-        this.collection
-      )
-      await this.fetchDocuments()
-      this.loading = false
     },
     performSearchDocuments,
     navigateToDocument() {
@@ -498,7 +498,7 @@ export default {
         size: this.paginationSize
       }
       try {
-        let searchQuery = null        
+        let searchQuery = null
         searchQuery = filterManager.toSearchQuery(this.currentFilter, this.collectionMapping)
         if (!searchQuery) {
           searchQuery = {}
--- a/src/components/Data/Documents/Page.vue
+++ b/src/components/Data/Documents/Page.vue
@@ -63,11 +63,11 @@
               </template>
             </b-col>
           </b-row>
-          <template v-if="fetchingDocuments">
+          <template v-if="loading">
             <b-row class="text-center">
               <b-col>
                 <b-spinner
-                  v-if="fetchingDocuments"
+                  v-if="loading"
                   variant="primary"
                   class="mt-5"
                 ></b-spinner>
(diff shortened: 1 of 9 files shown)


xbill82:
@morgandruesne GH checks didn't update, but tests are failing https://dashboard.cypress.io/projects/qnb41a/runs/1795

src/services/filterManager.ts
```diff
@@ -302,11 +302,11 @@ export const formatFromBasicSearch = (groups = [[]]) => {
 
       if (filter.operator === 'contains') {
         formattedFilter.bool.must.push({
-          regexp: { [filter.attribute]: '.*' + filter.value + '.*' }
+          regexp: { [filter.attribute]: '.*' + filter.value.toLowerCase() + '.*' }
```
xbill82, line 305: Interesting. I kinda remember we had the issue of case-sensitive Regex in Elasticsearch. Is it still the case?

scottinet:
What is "contains search"? Did you mean "content search"?

xbill82:
@morgandruesne I think that we should opt for a conditional query generation. 
If the field is of type `text`, we should use [match_phrase_prefix](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase-prefix.html).
if the field is of type `keyword`, we need to find out if ES provides something similar.

xbill82:
... something like this, probably https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-prefix-query.html

src/components/Data/Documents/Page.vue
```diff
@@ -423,8 +423,13 @@ export default {
     },
     // LIST (FETCH & SEARCH)
     // =========================================================================
-    loadAllTheThings() {
-      this.loadMappingInfo()
+    async loadAllTheThings() {
+      this.loading = true
+      try {
```
xbill82, line 428: using a try...catch block is a nice idea! I think you can put all the calls of the function inside it and display a toast rather than simply stop the loader

src/components/Data/Documents/Page.vue
```diff
@@ -493,8 +498,8 @@ export default {
         size: this.paginationSize
       }
       try {
-        let searchQuery = null
-        searchQuery = filterManager.toSearchQuery(this.currentFilter)
+        let searchQuery = null
```
xbill82, line 501:
```suggestion
        let searchQuery = null
```

src/services/filterManager.ts
```diff
@@ -301,13 +315,28 @@ export const formatFromBasicSearch = (groups = [[]]) => {
...
-          regexp: { [filter.attribute]: '.*' + filter.value.toLowerCase() + '.*' }
-        })
+        if (mappings[filter.attribute].type === "text") {
+          formattedFilter.bool.must.push({
+            match_phrase_prefix: { [filter.attribute]: filter.value }
+          })
+        }
+        if (mappings[filter.attribute].type === "keyword") {
+          formattedFilter.bool.must.push({
+            regexp: { [filter.attribute]: ".*" + buildCaseInsensitiveRegexp(filter.value) + ".*" }
```
xbill82, line 325: In order for the regex to behave like the `match_phrase_prefix` operator...
```suggestion
            regexp: { [filter.attribute]:  buildCaseInsensitiveRegexp(filter.value) + ".*" }
```

src/services/filterManager.ts
```diff
@@ -301,13 +315,28 @@ export const formatFromBasicSearch = (groups = [[]]) => {
...
-        })
+        
+        if (mappings[filter.attribute].type === "text") {
+          formattedFilter.bool.must_not.push({
+            match_phrase_prefix: { [filter.attribute]: filter.value }
+          })
+        }
+        if (mappings[filter.attribute].type === "keyword") {
+          formattedFilter.bool.must_not.push({
+            regexp: { [filter.attribute]: ".*" + buildCaseInsensitiveRegexp(filter.value) + ".*" }
```
xbill82, line 337: Same as above
```suggestion
            regexp: { [filter.attribute]: buildCaseInsensitiveRegexp(filter.value) + ".*" }
```

src/components/Data/Documents/Page.vue
```diff
@@ -423,23 +423,29 @@ export default {
...
+        filterManager.save(
+          this.currentFilter,
+          this.$router,
+          this.index,
+          this.collection
+        )
+        await this.fetchDocuments()
+        this.loading = false
+      } catch {
+        this.loading = false
```
xbill82, line 447: Toast here?

PR #721, opened 2020-06-04
CI test v1 and v2 backend

xbill82:
## What does this PR do ?
:point_up: This PR looks huge, but it's essentially due to the new package-lock.json. There's not so many files to review.

This PR refactors the CI to have end-to-end tests with v1 and v2 back-end. 
It also adds a new test runner based on OClif, providing more options ad a prettier output (featuring lovely emoji).

The tests are now split between single-backend and multi-backend. 
* Single-backend tests only spawn one Kuzzle instance, whose version is determined by the `--backend` option of the test-runner. The same spec files are played on the AC connected to both Kuzzle v1 and v2.
* Multi-backend test instantiate two Kuzzle (both v1, both v2 or one of each version) and tests the switch from one to another.

## TODO
- [x] rewrite the test runner as a single-command OClif
- [x] refactor the CI so that we have 2 different test jobs
- [x] make the V1 tests green
- [x] make the V2 tests green
- [x] implement a Cypress plugin to `doco up` and `doco down` the v1 and v2 stacks
- [x] e2e tests with the backend going down and getting back up again
- [x] e2e tests with the wrong backend version
- [x] e2e switching between different backends (same version, different version) 
- [x] separate single-backend and multi-backend specs
- [x] implement multi-backend option in test runner
- [x] add multi-backend test job in CI


src/components/Data/Documents/ViewMap.vue
```diff
@@ -39,7 +39,7 @@
             <li>
               <a
                 title="You are not allowed to delete this document"
-                :class="{ 'text-secondary': !canDelete }"
+                :class="{ disabled: !canDelete }"
```
xbill82, line 42: This is a little mistake I made while resolving conflicts with the `4-dev` branch. But this component is currently deactivated. I propose that we leave this as-is and we'll focus on the component once we restore it.

src/services/kuzzleWrapper-v1.ts
```diff
@@ -185,6 +221,139 @@ export class KuzzleWrapperV1 {
...
+    let bool: any = {}
+
+    bool.should = groups.map(filters => {
+      let formattedFilter: any = { bool: { must: [], must_not: [] } }
+      filters.forEach((filter: any) => {
+        if (filter.attribute === null) {
+          return
+        }
+
+        if (filter.operator === 'contains') {
```
Njuelle, line 261: Nitipicking: I think a `switch case` statement at least on `filter.operator` might improve readability.

src/services/kuzzleWrapper-v1.ts
```diff
@@ -14,13 +20,21 @@ let getValueAdditionalAttribute = (content, attributePath) => {
...
-  updater:
-    _kuzzle_info.updater === '-1' ? 'Anonymous (-1)' : _kuzzle_info.updater,
-  createdAt: _kuzzle_info.createdAt,
-  updatedAt: _kuzzle_info.updatedAt
-})
+
+function buildCaseInsensitiveRegexp(searchString) {
+  return searchString
+    .split('')
+    .map(e => {
```
Njuelle, line 27: Nitpicking: What does this `e` stand for? Is it for an error, an event or an element?

test/e2e/run-test/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+          )
+          const doco = execa('docker-compose', [
+            '-f',
+            docoFile,
+            '-p',
+            `stack-${version}`,
+            'up',
+            '-d'
+          ])
+          // doco.stderr.pipe(process.stderr)
```
Njuelle, line 71: if useless, remove it

xbill82: I just kept this since I often use it for debugging

PR #722, opened 2020-06-04
Json editor

### LINKED ISSUE #652: Make the raw JSON search field resizable
Make the height of the raw JSON search field resizable so that large queries are more readable.

berthieresteban:
## What does this PR do ?

fix #652 
Feature not included in the todo backlog of the beta release milestone but short to implement
Use the textarea of bootstrap-vue to be consistent with other ui implementations 
Little hack on the component to avoid `tab` click change the focus to the buttons and allow user to add double space using `tab` key

Many options can be use (check here https://bootstrap-vue.org/docs/components/form-textarea#form-textarea) like auto increase the height following the number of line typed, max lines displayed (scroll if there is more) etc

I arbitrarily chose 16 rows by default and let the possibility to change the size manualy using the bottom right corner of the textarea, please don't hesitate to say if you think of a better implementation

  - Step 1 : Run kuzzle
  - Step 2 : Mishandle every textarea input (document create/update, collection mapping, role, user, profile...)

### Other changes
/

### Boyscout
 - `collectionsCount` in `src/components/Data/Leftnav/IndexBranch.vue` 

image


xbill82:
We use ACEeditor to provide syntax and error highlighting while writing JSON. Replacing it with a textarea would not be a good improvement, IMHO. I don't think it is a good idea going forward with this PR.

berthieresteban:
I understood this and I'm working on another implementation keeping the jsonEditor currently used to add the possibility to resize

PR #738, opened 2020-06-16
[#736] Force editing malformed environement

### LINKED ISSUE #736: Previous environments are not compatible
## Expected Behavior

I can use my previous environments

## Current Behavior

I can't
image

## Possible Solution

Display a message to tell the users to choose a Kuzzle version for their environments

## Steps to Reproduce

Create an environment in console.kuzzle.io, export it and then import it on the new version (the one supporting kuzzle v1 and v2)

Njuelle:
## What does this PR do ?
Fix #736 
As discussed, since it is impossible for us to automatically select the right version of the SDK depending on the environment and since we cannot force a default version. The proposed solution is to signal to the end user the environments that are malformed and when clicking on one of these environments force the opening of the editing modal.

image

image

### Boyscout
* fix failing update user mapping test

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,6 @@
 <template>
   <div class="App">
-    <main-spinner v-if="initializing"></main-spinner>
-    <template v-else>
+    <template>
       <router-view
         data-cy="App-loggedIn"
         @environment::create="editEnvironment"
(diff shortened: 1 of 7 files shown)


berthieresteban:
We might add a warning icon on the dropdown too 
Here with old environment it just loop and display connecting but we don't now something is wrong
image

berthieresteban:
After editing an old environment to add a backendMajorversion we must reload the page to be able to connect to the backend otherwise we are stuck on the same screenshot of the previous comment

src/App.vue
```diff
@@ -41,48 +40,19 @@ import {} from './assets/global.scss'
...
+    ModalImport
   },
   data() {
     return {
-      initializing: true,
       environmentId: null
     }
   },
-
-  async mounted() {
```
xbill82, line 61: * This must be restored
* The route guard must be deleted
* we need to add to this the redirection to `CreateEnvironment` in case no environments are present after the loading

xbill82:
Tests are finished but the result hasn't reached GH

PR #754, opened 2020-06-22
Add color state verification when editing an environment

### LINKED ISSUE #753: Colors verification is missing in environment management
## Expected Behavior
be informed that the selection of a color is missing when creating or editing an environment

## Current Behavior
the form is valid even though the color is missing.

## Context (Environment)
Admin Console Version: 4.x

Njuelle:
## What does this PR do ?
Fix #753 
Add color state verification when editing an environment

image

### Other changes
* shows version feedback only when the version is missing

### DIFF
--- a/src/components/Common/Environments/CreateEnvironment.vue
+++ b/src/components/Common/Environments/CreateEnvironment.vue
@@ -233,8 +233,8 @@ export default {
       if (this.submitting) {
         return true
       }
-      // need ternary conditon because version can be null
-      return this.environment.backendMajorVersion ? true : false
+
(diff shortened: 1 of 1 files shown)


Aschen:
I don't understand why the input stay green after selecting an environment? 
IMHO it's important to highlight it in red until the user input the environment but after we shouldn't have a visual feedback unless there is some other information (eg: try to detect Kuzzle v2 and put the input in red if it's Kuzzle v1)

Njuelle:
@Aschen, you're right, it was a little side effect. It's fixed now ;)

PR #778, opened 2020-07-06
Vertical layout improvements

### LINKED ISSUE #774: Textarea for editing document is too short and mapping is automatically unfolded
Being able to edit more than 15 rows of content

## Current Behavior

Small textarea and veryyy long page with the mapping

## Possible Solution

Let the mapping folded and unfoldable + let the user adjust (and save  the setting?) the textarea if possible

## Steps to Reproduce

1.
2.
3.
4.

## Context (Environment)

Kuzzle version:
Node.js version:
Admin Console Version:

xbill82:
## What does this PR do ?
Fixes #774 by allowing the JSON editor and the column on the right to take all the available height but not more. This improvement has been applied to all the forms that contain a JSON editor.

output

Open one of the following forms
* Create/Update document
* Create/Update collection
* Create/Update role
* Create/Update profile
* Edit user custom mapping
and resize your window vertically.

### DIFF
--- a/src/components/Security/Profiles/CreateOrUpdate.vue
+++ b/src/components/Security/Profiles/CreateOrUpdate.vue
@@ -2,7 +2,7 @@
   <b-card class="ProfileCreateOrUpdate">
     <!-- Json view -->
     <b-row class="h-100">
-      <b-col lg="7" md="12" class="ProfileCreateOrUpdate-editorLayout">
+      <b-col lg="7" md="12" class="d-flex flex-column">
         <b-form-group
           v-if="!id"
           label="Profile ID"
@@ -35,7 +35,7 @@
       </b-col>
 
       <!-- Mapping -->
-      <b-col lg="5" md="12" class="ProfileCreateOrUpdate-cheatsheetLayout">
+      <b-col lg="5" md="12" class="d-flex flex-column">
         <h3>Cheatsheet</h3>
         <div class="ProfileCreateOrUpdate-cheatsheet">
           Your profile is a set of <code>policies</code>, each of which will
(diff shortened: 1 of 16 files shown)


src/components/Security/Profiles/CreateOrUpdate.vue
```diff
@@ -97,6 +100,27 @@
   </b-card>
 </template>
 
+<style lang="scss" scoped>
+.ProfileCreateOrUpdate {
+  // margin-bottom: 3em;
```
Njuelle, line 105:
```suggestion
```

src/components/Security/Profiles/Update.vue
```diff
@@ -14,13 +14,21 @@
   </b-container>
 </template>
 
+<style lang="scss" scoped>
```
Njuelle, line 17: IMHO we can use Bootstrap classes: `d-flex flex-column h-100`
Let's keep the custom classes for when we really can't do without

src/components/Security/Roles/CreateOrUpdate.vue
```diff
@@ -97,6 +99,28 @@
   </b-container>
 </template>
 
+<style lang="scss" scoped>
```
Njuelle, line 102: Styles duplicated with that of the `profiles/createOrUpdate` component.
Though I don't know if it's a good thing to refactor it here and now

src/components/Security/Users/EditCustomMapping.vue
```diff
@@ -83,11 +75,19 @@
           </div>
         </template>
       </b-card>
-    </div>
+    </template>
   </b-container>
 </template>
 
 <style lang="scss" scoped>
+.EditUserMapping {
```
Njuelle, line 83: Also use bootstrap classes here

PR #783, opened 2020-07-09
Form validation w/Vuelidate

xbill82:
## What does this PR do ?
Leverages Vuelidate for form validation in the following components

* `src/components/Common/Environments/CreateEnvironment.vue`
* `src/components/Common/Login/ResetPasswordForm.vue`
* `src/components/Data/Collections/CreateOrUpdate.vue`
* `src/components/Security/Profiles/CreateOrUpdate.vue`
* `src/components/Security/Roles/CreateOrUpdate.vue`
* `src/components/Security/Users/CreateOrUpdate.vue`

Open the form and type some invalid input. Check that the visua feedback is shown and that you cannot submit the form.

### Boyscout
* Cleaned-up useless styles
* Reworded some messages
* Fixed bug in update profile


src/components/Common/Environments/CreateEnvironment.vue
```diff
@@ -102,60 +108,88 @@
...
 
 <script>
-import { DEFAULT_COLOR, envColors } from '../../../vuex/modules/kuzzle/store'
+import { validationMixin } from 'vuelidate'
+import { numeric, required } from 'vuelidate/lib/validators'
+import { isValidHostname } from '../../../utils'
+
+import { envColors, DEFAULT_COLOR } from '../../../vuex/modules/kuzzle/store'
 const useHttps = window.location.protocol === 'https:'
+function nameIsUnique(value) {
```
Njuelle, line 129: Why declare these methods here? Can't we put them in the `methods` object?

src/components/Data/Collections/CreateOrUpdate.vue
```diff
@@ -150,11 +154,24 @@
...
 
 <script>
+import { validationMixin } from 'vuelidate'
+import { requiredUnless } from 'vuelidate/lib/validators'
+
 import Headline from '../../Materialize/Headline'
 import Focus from '../../../directives/focus.directive'
 import JsonEditor from '../../Common/JsonEditor'
 
+function isValidCollectionName(value) {
```
Njuelle, line 164: same here, maybe we can put this function in the `methods` object

src/components/Common/Environments/CreateEnvironment.vue
xbill82: 1. these little functions are meant to be generic and pure
2. it feels wonky to declare your `validations` object by referencing `this.function`, since you don't really know if Vue first declares the methods or the validation system
3. vuelidate dynamically changes the closure around the custom validators when they are called. If the component scope is needed, it will be dynamically injected.

src/components/Data/Collections/CreateOrUpdate.vue
xbill82: same as above

test/e2e/cypress/integration/single-backend/users.spec.js
```diff
@@ -108,12 +108,11 @@ describe('Users', function() {
       .type('{selectall}{backspace}', { delay: 200, force: true })
       .type(
         `{
-"query": { 
+"query": {
```
alexandrebouthinon, line 111: Wrong indentation

xbill82: Indentation here is not added since the ACE editor automatically adds it while Cypress types the text inside it.

Njuelle (approved):

berthieresteban:
Great job ! 
Feels good to have a clean error handle and feedback :heart:

Here are some points: 
  - It should be good to add vuelidate in the index creation too (actually just type a space, ok will turn enable but click it doesn't display any error and don't do anything)
  - Also in the index creation the tip say "The index name should contain only lowercase characters, no space and cannot begin with an underscore (_)" but we can create an index starting with an "_" 
  - User creation: I've seen a weird  behaviour, I created an user with admin profile, no kuid provided nor credentials, I try to do that a second time but a toast appears and said that I must add an username, same toast when i tried to update the first created account (without credentials)

src/components/Data/Indexes/CreateIndexModal.vue
```diff
@@ -15,26 +15,47 @@
...
         label="Index name"
         label-for="indexName"
-        description="The index name should contain only lowercase characters, no space and cannot
-          begin with an underscore (_)"
       >
+        <template v-slot:description
+          >The index name should contain only lowercase characters and no
+          spaces. It also must not contain de following characters:
+          <code>\</code>, <code>/</code>, <code>*</code>, <code>?</code>,
+          <code>"</code>, <code>&lt;</code> <code></code>, <code>></code>,
```
berthieresteban, line 33:
```suggestion
          <code>"</code>, <code>&lt;</code>, <code>></code>,
```

src/components/Data/Indexes/CreateIndexModal.vue
```diff
@@ -15,26 +15,47 @@
...
+          <code>|</code>, <code>,</code>, <code>#</code>, <code>:</code>,
+          <code>%</code>, <code>&</code>, <code>.</code> "</template
```
berthieresteban, line 35:
```suggestion
          <code>%</code>, <code>&</code>, <code>.</code> </template
```

berthieresteban (approved):

PR #786, opened 2020-07-10
Csv export

berthieresteban:
## What does this PR do ?
#787 

Allow user to export a csv file of the current column view (only the current page with the selected columns and the search query)

  - Step 1 : create a testIndex index, a testCollection collection, some documents with
  - Step 2 : Check the column view and select some property to display
  - Step 3 : Click on the export button
  ...

### Other changes
/

### Boyscout
/
/


Aschen:
IMHO this feature should allows to export every documents matching the request instead of only the current page

berthieresteban:
@Aschen 
You're right, it might be an evol because it will probaby need an UX reflexion.
Actually we can export the array we are seeing on the column view wich is very clear IMHO

src/components/Data/Documents/Views/Column.vue
```diff
@@ -333,6 +343,32 @@ export default {
     }
   },
   methods: {
+    exportFilename() {
```
Njuelle, line 346: better to use a computed here

src/components/Data/Documents/Views/Column.vue
```diff
@@ -333,6 +343,32 @@ export default {
...
+      const date = new Date()
+      return `${this.index}_${this.collection}_${date.getMonth() +
+        1}-${date.getDate()}-${date.getFullYear()}.csv`
+    },
+    csvExport() {
+      const documents = []
+      for (const item of this.formattedItems) {
+        const itemArray = []
+        delete item.actions
```
Njuelle, line 355: You can use `_.omit()` here

xbill82:
⚠️  **TL;DR. Stop this feature please** ⚠️ 

Sorry to be the party-dumper here, but... I think we should freeze this PR before going any further.
I mean, who wants to export the current page? Really... Who would be happy to see this feature enabling to export _just_ the current page without running to our Discord to ask for the _real_ feature, i.e. exporting the whole result set?
Nobody.
When you export, you export the data, not the view. The current page is just a view for the whole result set.
The feature here is "export the search results", which is doable. We need to think about some details before going on (that's why I'd like to freeze).

* The UX of the workflow (button, color, layout, etc... but looks pretty easy)
* The memory limitations of the client, which is the ugly part. Again: who does a full export of a production base _from the browser_? Because if we make this button available, you can be sure someone will try to dump their whole prod base with it. And prod bases can be huge and bloat the client. The problem here is, in order to generate a CSV file from the browser, you need to load all the data in memory (or worse, in the DOM), then tell the browser to dump them in a file for download. Kourou does it in a much smarter way: it writes the data in a file, as long as it is received from the server. And it can do it this way because it runs in a Nodejs environment where the `fs` API is available (which is not in the browser). This is why I'd keep this feature _out of the browser environments_. But we can mitigate this statement by adding things like result set size limitations. We might imagine to export the result set unless it is bigger than `X` result. What is `X`? Dunno. I'm unable to estimate it.

Any thoughts?

Aschen:
@xbill82 We can stream the CSV line per line just as we wrote the JSONL line per line in a file with Kourou.
So we need to use the stream API but this feature is far than easy IMHO

(And I agree, as I say in my first comment, this feature need to be able to export the whole data set)

xbill82:
@Aschen steaming the CSV would be OK, but can we do it directly to a file? I think this is forbidden in a browser. I mean, the user will have to load _in memory_ the whole result-set, which can have hard consequences on the stability of the application depending on the size of the result-set.

PR #793, opened 2020-07-22
fix users not displayed

### LINKED ISSUE #767: Users are not displayed on the users page
## Expected Behavior

I should be able to see users list in the "Users" page

## Current Behavior

Users are not listed
image

## Steps to Reproduce

 - Creates an user with the admin console
 - Load the users list page

## Context (Environment)

Kuzzle version: 2.3.0
Admin Console Version: next-console.kuzzle.io (v4)

berthieresteban:
## What does this PR do ?
Fix #767 users not displayed when url contains from=XXX and there are less than XXX users in kuzzle by forcing AC to set from=0

  - Step 1 : Create an user
  - Step 2 : go to `http://localhost:8080/#/security/users?from=10`
  - Step 3 : check if from=10 disappear from url and user is displayed

### Other changes
/

### Boyscout
/

### DIFF
--- a/test/e2e/cypress/integration/single-backend/users.spec.js
+++ b/test/e2e/cypress/integration/single-backend/users.spec.js
@@ -115,7 +115,7 @@ describe('Users', function() {
 "must": {
 "match_phrase_prefix": {
 "name": "Dummy User (${kuids[1]})"{downarrow}{downarrow}{downarrow}{downarrow}
-}`,
+`,
         {
           force: true
         }
--- a/test/e2e/cypress/integration/single-backend/users.spec.js
+++ b/test/e2e/cypress/integration/single-backend/users.spec.js
@@ -110,7 +110,7 @@ describe('Users', function() {
       .type('{selectall}{backspace}', { delay: 200, force: true })
       .type(
         `{
-"query": { 
+"query": {
 "bool": {
 "must": {
 "match_phrase_prefix": {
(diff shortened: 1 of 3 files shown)


berthieresteban:
@xbill82 I made a conditionnal test for this bug-fix since this pr only fix for kuzzle v2.
I'm not sure about how I did it, could you check that please ?
For example, if the Cypress.env('BACKEND_VERSION') isn't an integer the test will pass for both kuzzle version...

xbill82:
@berthieresteban nice point. I think the env variable is always a String.

test/e2e/cypress/integration/single-backend/users.spec.js
```diff
@@ -284,11 +284,36 @@ describe('Users', function() {
...
     cy.get(
       '[data-cy="UserManagement-pagination"] .page-link[aria-posinset="2"]'
-    ).click({force: true})
+    ).click({ force: true })
     cy.get('[data-cy=UserItem]').should('have.length', 4)
     cy.url().should('contain', 'from=10')
   })
 
+  it('Should be able to lists the users with a wrong from url parameter', () => {
+    if (Cypress.env('BACKEND_VERSION') === 2) {
```
xbill82, line 293: Instead of this, you should actually use `cy.skipOnBackendVersion(1)` that @Njuelle developed

xbill82 (approved):

Njuelle (approved):

PR #796, opened 2020-08-06
Fix in the display of range Input

morgandruesne:
## What does this PR do ?
This RP modifies the cols of Advanced Filter range to make it more human readable
also adds a popover to explain how a attribute appears in the attribute list

image


xbill82:
Please, apply this fix to restore the Netlify build https://github.com/kuzzleio/kuzzle-admin-console/pull/793/commits/d5f9a5245757776e074767c457a8e227580c82bd

PR #813, opened 2020-08-26
Fix re-authenticate when token expires

morgandruesne:
## What does this PR do ?
Change the e.id to allow the user to re-authenticate after the token has expired


src/components/Home.vue
```diff
@@ -70,14 +70,16 @@ export default {
   },
   methods: {
     onTokenExpired() {
+      console.log("plop");
```
xbill82, line 73:
```suggestion

```

xbill82 (approved):

test/e2e/cypress/integration/single-backend/login.spec.js
```diff
@@ -52,7 +52,7 @@ describe('Login', function() {
     cy.get('[data-cy="App-loggedIn"]')
   })
 
-  it('Should be able to login without losing context when token expires', () => {
+  it.only('Should be able to login without losing context when token expires', () => {
```
Njuelle:
```suggestion
  it('Should be able to login without losing context when token expires', () => {
```

Njuelle (changes requested):

Njuelle (approved):

Issue #814, opened 2020-09-01
QuickFilter : Bad alignement on some screens resolutions


Njuelle:
Bad alignement on some screens resolutions for the QuickFilter input

<img width="1204" alt="Capture d’écran 2020-09-01 à 11 58 38" src="https://user-images.githubusercontent.com/3192870/91837227-34756580-ec4c-11ea-9b79-8525d6fb38d6.png">

PR #820, opened 2020-09-07
#814 Fix quick filter input alignment

Njuelle:
# What does this PR do ?
wrap and limit QuickFilter input width

quickFilter input


xbill82 (commented):
Why adding a prop and limiting to 75% instead of wrapping the component into a parent that forces the width?

Njuelle:
The idea was not to add more CSS but after discussions this seems to be the simplest solution. I will modify this PR.

xbill82 (approved):

berthieresteban (approved):

PR #822, opened 2020-09-09
Make BasicFilter more responsive

morgandruesne:
## What does this PR do ?
this PR makes the basic filter more responsive, so more pleasant and readable on smaller screens.

Normal size : 
image

Screen < 1140px :
image

Also, fix bug of `isRealtime` not exit in `DropdownAction.vue `

### DIFF
--- a/src/components/Common/Filters/BasicFilter.vue
+++ b/src/components/Common/Filters/BasicFilter.vue
@@ -109,7 +109,7 @@
                 />
               </template>
             </b-col>
-            <b-col xl="1" class="mb-1 text-center">
+            <b-col xl="1" class="mb-1 text-right">
               <i
                 class="fa fa-times mt-2 pointer"
                 @click="removeAndBasicFilter(groupIndex, filterIndex)"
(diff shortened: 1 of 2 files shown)


berthieresteban:
Seems nice !
I just think the cross button to remove a condition isn't really visually explicit and too close from the "and" text or button on the second screenshot

xbill82:
Nice one! I agree with Esteban! I'd try to put the cross on the right side of the condition (there is plenty of space)

berthieresteban:
Better like that :)
But again I'm not sure it's enough explicit
On the first screen we understand easily that the cross will remove the entire line
On the second not sure that is the same way because the condition is a sort of bloc on multiple lines
What do you think about an outlined red button with "Remove Condition" ? @xbill82 @morgandruesne

morgandruesne:
With the remove button on the left it looks more natural. What do you think?

xbill82:
I've tried a suggestion, what do you think?

93089835-2ee03c80-f69c-11ea-9bad-9c6d11fb4389

xbill82 (approved):
Looks clear enough 😄

Leodau (approved):

PR #826, opened 2020-09-14
Refactor Index/collection vuex store

Njuelle:
## What does this PR do ?

Add lazy loading for fetching index and collection list to avoid long loading times (several minutes sometimes). This change is accompanied by a complete refactor of the index/collection store to make the next evolutions easier. In addition, all the components of the "Data" section as well as the router have been adapted for the new behavior of the store.

From now on, the actions of index and collection retrieval are located only in the store, moreover they are called by the "DataLayout" component which is in charge of conditioning these calls according to the parameters coming from the router like this:

* Fetch the index list
* If the parameter `indexName` is present: fetch the collection list for the given index
* If the `collectionName` parameter is present: fetch the mapping for the given collection  

Child components are only displayed when these actions are completed.

### Other changes
Add new bulk delete indexes and collections modal : Instead of having to confirm the deletion for each selected item, just type "DELETE" once to delete everything


src/components/Data/Indexes/Page.vue
```diff
@@ -126,42 +129,45 @@
...
-    <CreateIndexModal :id="createIndexModalId" />
+    <CreateIndexModal
+      :id="createIndexModalId"
+      @modal-close="onCreateModalClose"
+    />
     <DeleteIndexModal
       :id="deleteIndexModalId"
       :index="indexToDelete"
-      @modal-close="onModalClose"
+      @modal-close="onDeleteModalClose"
```
xbill82, line 170: BootstrapVue exposes a @hide event, right for this purpose, if I'm not mistaken.

src/components/Data/Indexes/Page.vue
```diff
@@ -223,69 +229,80 @@ export default {
...
+          }
+        )
       }
     },
-    allChecked() {
-      if (!this.selectedIndexes || !this.filteredIndexes) {
-        return false
+    deleteIndexes() {
+      if (this.bulkDeleteEnabled) {
+        this.openDeleteModal(this.selectedIndexes[0])
```
xbill82, line 281: Why showing the modal only if `bulkDeleteEnabled`? The usual way I do deletion with optional bulk is
* use a `itemsToDelete` array in `data()`
* the modal takes `itemToDelete` as a prop
* if `itemsToDelete.length === 1`, the modal asks the name of the item to confirm deletion
* else, the modal just shows a warning message before deleting

I don't really see the use for a boolean telling whether we're bulk deleting or not.

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
+  indexes: [],
+  loadingIndexes: false
 }
 
+const INTERNAL_INDEX_NAME = '%kuzzle'
+
 const mutations = createMutations<IndexState>()({
   reset(state) {
-    state.indexesAndCollections = {}
+    state.indexes = []
```
xbill82, line 26:
```suggestion
    state.indexes = Index[]
```

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
+  CreateCollectionPayload
+} from './types'
 import { createMutations, createModule, createActions } from 'direct-vuex'
 import { moduleActionContext } from '@/vuex/store'
 
 const state: IndexState = {
-  indexesAndCollections: {},
-  loadingIndexes: false,
-  loadingCollections: false
+  indexes: [],
```
xbill82, line 18:
```suggestion
  indexes: Index[],
```

src/vuex/modules/index/types.ts
```diff
@@ -1,11 +1,110 @@
...
+  public addCollection(collection: Collection) {
+    if (!this.collections) {
+      throw new Error(
+        'Unable to perform operations, the collection list is not yet initialized'
+      )
+    }
+    this.collections.push(collection)
+  }
+
+  public removeCollection(collection: Collection) {
```
xbill82, line 31: Warning.
You are doing OOP in the store, which is meant to be mutation-oriented. Such an operation will not be tracked by the Vuex DevTools, since it is not performed by a mutation. I really like the OOP-style here, but it is breaking the paradigm, which is the only thing that prevents us from getting crazy on the store (which is a global object, let's not forget it).

xbill82: Maybe you could try to call this method within a mutation, instead of just calling it directly. This way, the benefit of traceability and time-travel provided by Vuex is kept together with the OOP-style separation of concerns.

src/vuex/modules/index/types.ts
```diff
@@ -1,11 +1,110 @@
...
+
+  get name() {
+    return this._name
+  }
+
+  get CollectionsCount(): number | undefined {
+    return this.collections?.length
+  }
+
+  public initCollections(collections: Collection[]) {
```
xbill82, line 18: About putting logic in the types.

It is very appealing to do it this way, since we all like to properly separate the code. But beware: since this code is no longer in a mutation, it can now be called outside a mutation, which will break the Vuex paradigm (which, let's not forget it, is the only thing that makes a global object like the store viable). In the current implementation, the logic is always called by the mutations (I think, I hope) but... Are we sure we will always remember to do so?
According to Murphy's Law, the answer is no.

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
     let result = await rootGetters.kuzzle.$kuzzle.index.list()
-    result = result.filter(index => index !== '%kuzzle')
+    result = result.filter(
+      (indexName: string) => indexName !== INTERNAL_INDEX_NAME
+    )
 
-    for (const index of result) {
-      commit.addIndex(index)
+    for (const indexName of result) {
+      // we only need to add the indexes not yet fetched
```
xbill82, line 92: What about indexes that might have been deleted?

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
-    })
+    const result = await rootGetters.kuzzle.$kuzzle.collection.list(index.name)
 
-    commit.setLoadingCollectionsForIndex({ index, loading: false })
-  },
+    const collections = result.collections.array.map(el => {
+      const collectionType =
+        el.type === CollectionType.STORED
+          ? CollectionType.STORED
+          : CollectionType.REALTIME
```
xbill82, line 110: This implies that a collection of an unknown type (like, in the future) will be treated as a `REALTIME` one. You should use this check to trigger an exception if the type is unknown.

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
-      collections.push({ index: index, collection: collection })
-      localStorage.setItem('realtimeCollections', JSON.stringify(collections))
-      commit.addRealtimeCollection({ index: index, name: collection })
-      return Promise.resolve()
-    }
+    commit.setLoadingCollectionsForIndex({ index, loading: true })
+
+    const collectionType = isRealtime
+      ? CollectionType.STORED
+      : CollectionType.REALTIME
```
xbill82, line 140: Same as above, maybe this check should be factorized into a separate function?

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,172 @@
...
-    ) {
-      dispatch.removeRealtimeCollection({ index, collection })
+  async deleteCollection(
+    context,
+    { index, collection }: IndexCollectionPayload
+  ) {
+    const { commit, rootGetters } = indexActionContext(context)
+    commit.setLoadingCollectionsForIndex({ index, loading: true })
+
+    if (index.doesCollectionExist(collection.name)) {
```
xbill82, line 161: Else? What else... ☕

src/vuex/modules/index/types.ts
xbill82: What can we do to prevent Murphy's Law or, at least, mitigate its prophecy? Ideas...
* Giving up the OOP-style stuff and just putting the logic in the mutations (after all, Vuex/Redux/Flux paradigms are all about functional programming, but I know this is not the kind of things you want to read at this point).
* Writing JSDoc comments on _all_ state-mutating methods that would _try to_ warn the developer about not calling these methods outside a mutation.
* Adding some weird `fn.caller` logic that would trigger an error if the method is not called by a mutation (dunno how to do this).

😬

xbill82 (commented):
👏

src/components/Common/Breadcrumb.vue
```diff
@@ -143,22 +143,19 @@
 export default {
   name: 'CommonBreadcrumb',
   methods: {
+    index() {
+      return this.$route.params.index
+        ? this.$store.direct.getters.index.getOneIndex(this.$route.params.index)
```
xbill82, line 148: Suggestion: un-bloat your code by using `mapGetters`

src/components/Data/Collections/Create.vue
```diff
@@ -22,31 +22,32 @@ export default {
...
-            mapping: payload.mapping,
-            dynamic: payload.dynamic
-          })
-          .then(() => {
-            this.$router.push({
-              name: 'Collections',
-              params: { index: this.index }
-            })
-          })
+        await this.$store.direct.dispatch.index.createCollection({
```
xbill82, line 39: Same here, you can use `mapActions` to un-bloat this long call (I never use it bc I just discovered it...) https://vuex.vuejs.org/api/#mapactions

src/components/Data/Indexes/Page.vue
```diff
@@ -106,7 +106,7 @@
             />
           </template>
           <template v-slot:cell(collectionCount)="row">
-            <span>{{ row.item.collectionCount || '--' }}</span>
+            <span>{{ row.item.collectionsCount || '--' }}</span>
```
xbill82, line 109:
```suggestion
            <span>{{ row.item.collectionCount || '--' }}</span>
```

src/vuex/modules/index/getters.ts
```diff
@@ -1,9 +1,27 @@
...
-export const getOneByName = state => (indexName: string) => {
+export const getOneIndex = state => (indexName: string) => {
   return state.indexes.find(el => el.name === indexName)
 }
 
 export const indexes = (state: IndexState) => state.indexes
 
 export const loadingIndexes = (state: IndexState) => state.loadingIndexes
+
+export const loadingCollections = state => (indexName: string) => {
```
xbill82, line 12:
```suggestion
export const loadingCollections = (state: IndexState) => (indexName: string) => {
```

src/vuex/modules/index/store.ts
```diff
@@ -145,40 +158,113 @@ const actions = createActions({
...
+    const collectionType = isRealtime
+      ? CollectionType.STORED
+      : CollectionType.REALTIME
+
+    let updatedCollection = new Collection(name, collectionType)
+
+    updatedCollection.mapping = mapping
+    updatedCollection.dynamic = dynamic
+
+    await rootGetters.kuzzle.$kuzzle.query({
```
xbill82, line 199: Why do we use `query` here? Is this compatible btw v1 and v2?

src/components/Common/Breadcrumb.vue
Njuelle: Yes, I know mapGetter but finally, in use, I find it's simpler and faster to use the global store object (no import to add, syntax easier to remember).

src/vuex/modules/index/store.ts
Njuelle: I had first tested with the dedicated method of the SDK but for some reason it didn't work (maybe related to `kuzzleWrapper`). Since it works very well with `query`, my plan is to add another TODO and deal with the problem in another PR

src/components/Data/Collections/ModalClear.vue
```diff
@@ -71,9 +71,12 @@ export default {
       }
 
       try {
-        await this.$store.direct.dispatch.collection.clearCollection({
+        await this.$kuzzle.query({
```
xbill82, line 74: ⚠️ this is not meant to work. Kuzzle SDK is now accessed via the store getter, which returns the version corresponding to the current environment major version. If `this.$kuzzle` works it means there's something weird going on.

src/components/Data/Collections/CollectionList.vue
```diff
@@ -365,23 +365,23 @@ export default {
       )
     },
     async onDeleteModalSuccess() {
-      await this.fetchCollections()
+      await this.fetchCollectionsList()
```
xbill82, line 368:
```suggestion
      await this.fetchCollectionList()
```

In English, when you have a list of items you say "Item List", not "Items List" (the object of the list is always singular, since the List expresses the plurality).

src/components/Data/Indexes/Page.vue
```diff
@@ -271,7 +271,7 @@ export default {
     },
     async refreshIndexes() {
       try {
-        await this.$store.direct.dispatch.index.listIndexes()
+        await this.$store.direct.dispatch.index.fetchIndexesList()
```
xbill82, line 274:
```suggestion
        await this.$store.direct.dispatch.index.fetchIndexList()
```
Same as above

src/components/Data/Collections/ModalClear.vue
Njuelle: indeed it doesn't work, a small error of copy/paste

src/components/Data/Layout.vue
```diff
@@ -19,27 +19,120 @@
...
+  data() {
+    return {
+      isFetching: false
+    }
+  },
   computed: {
     loading() {
-      return this.$store.direct.state.index.loadingIndexes
+      return (
+        this.isFetching ||
```
xbill82, line 40: I wonder if this flag is really necessary

Njuelle: in case the loading state has not yet started (it depends on which action is performed first) and the child components try to render while the data is not yet present

src/vuex/modules/index/store.ts
Njuelle: ?

src/components/Data/Documents/Page.vue
```diff
@@ -696,7 +708,7 @@ export default {
         }
       }
 
-      findDateFields(this.collectionMapping, null)
+      findDateFields(this.collectionNameMapping, null)
```
xbill82, line 711: Isn't this a replaceAll error?

Njuelle: oops, well seen

src/components/Data/Documents/Page.vue
```diff
@@ -498,7 +517,7 @@ export default {
     },
     async fetchDocuments() {
       this.$forceUpdate()
-      this.indexOrCollectionNotFound = false
+      this.indexNameOrCollectionNotFound = false
```
xbill82, line 520: 🤔 Wrong search & replace?

src/components/Data/Documents/Page.vue
```diff
@@ -541,7 +561,7 @@ export default {
       } catch (e) {
         this.$log.error(e)
         if (e.status === 412) {
-          this.indexOrCollectionNotFound = true
+          this.indexNameOrCollectionNotFound = true
```
xbill82, line 564: same as above?

src/components/Data/Documents/Page.vue
```diff
@@ -666,7 +678,7 @@ export default {
       this.$router.push({ query: mergedQuery }).catch(() => {})
     },
     addHumanReadableDateFields() {
-      if (!this.collectionMapping) {
+      if (!this.collectionNameMapping) {
```
xbill82, line 681: 🤔 ctrl + f?

src/components/Data/Indexes/DeleteIndexModal.vue
```diff
@@ -1,40 +1,47 @@
...
     size="lg"
-    :id="id"
-    @hidden="resetForm"
+    :id="modalId"
+    :ref="modalId"
+    @hide="resetForm"
   >
     <template v-slot:modal-title>
-      Index <strong>{{ truncateName(index) }}</strong> deletion
+      <template>
```
xbill82, line 10: What is this `<template>` for?

src/components/Data/Layout.vue
```diff
@@ -2,45 +2,187 @@
...
+        }
+
+        await this.$store.direct.dispatch.index.fetchCollectionList(index)
+      } catch (error) {
+        this.$log.error(error)
+        this.$bvToast.toast(
+          'The complete error has been printed to the console.',
+          {
+            title:
+              'Ooops! Something went wrong while fetching the collections list.',
```
xbill82, line 106:
```suggestion
              'Ooops! Something went wrong while fetching the collection list.',
```

src/vuex/modules/index/store.ts
xbill82: Sorry, I don't remember why we didn't apply this...

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,269 @@
...
-      !state.indexesAndCollections[index] ||
-      !state.indexesAndCollections[index].stored
-    ) {
-      return
-    }
-
-    const idx = state.indexesAndCollections[index].stored.indexOf(collection)
-    Vue.delete(state.indexesAndCollections[index].stored, idx)
+  removeIndexes(state, indexes: Index[]) {
+    const keepedIndexes = _.difference(state.indexes, indexes)
```
xbill82, line 73:
```suggestion
    const keptIndexes = _.difference(state.indexes, indexes)
```

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,269 @@
...
+  },
+  async fetchIndexList(context) {
+    const { commit, rootGetters } = indexActionContext(context)
+    const indexes: Index[] = []
     commit.setLoadingIndexes(true)
+
     let result = await rootGetters.kuzzle.$kuzzle.index.list()
-    result = result.filter(index => index !== '%kuzzle')
+    result = result.filter(
+      (indexName: string) => indexName !== INTERNAL_INDEX_NAME
```
xbill82, line 113: I was wondering if this was still relevant... Kuzzle does not return the internal indexes anymore, does it?

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,269 @@
...
-    commit.reset()
-    await dispatch.listIndexes()
+    if (!name) {
+      throw new Error('Invalid collection name')
+    }
 
-    for (let index in state.indexesAndCollections) {
-      await dispatch.listCollectionsForIndex(index)
+    if (index.doesCollectionExist(name)) {
+      throw new Error(`Collection "${name}" already exist`)
```
xbill82, line 151:
```suggestion
      throw new Error(`Collection "${name}" already exists`)
```

src/vuex/modules/index/store.ts
```diff
@@ -1,230 +1,269 @@
...
+    context,
+    { index, collection }: IndexCollectionPayload
+  ) {
+    const { commit, rootGetters } = indexActionContext(context)
+
+    if (!index.doesCollectionExist(collection.name)) {
+      throw new Error(`Collection "${collection.name}" doesn't exist`)
     }
+
+    commit.setLoadingCollections({ index, loading: true })
```
xbill82, line 255: 🤔 Do we really need this in this action?

src/vuex/modules/index/types.ts
```diff
@@ -1,11 +1,142 @@
...
+
+  get type(): CollectionType {
+    return this._type
+  }
+
+  public isRealtime(): boolean {
+    return this._type === CollectionType.REALTIME
+  }
+
+  private findType(name: string, type: string) {
```
xbill82, line 92: I would have created a `enum` type, rather than making all these checks, but if you wanna keep it like this...
```suggestion
  private checkType(name: string, type: string) {
```

xbill82: oh... I see the `enum` right below... Doesn't the language check the enum for you?

xbill82, line 92: Oh! The language does not type check enums??

src/vuex/store.ts
```diff
@@ -15,7 +14,6 @@ const { store, rootActionContext, moduleActionContext } = createDirectStore({
   modules: {
     auth,
     kuzzle,
-    collection,
```
xbill82, line 18: 🎉

xbill82 (approved):
SUPA DOOPA ROOPA WOOPA TROOPA CHOOPA KOOPA MEGA GIGA TERA FREAKIN GOOD JOB DUDE
💪 🎉

src/vuex/modules/index/store.ts
Njuelle: Because the `IndexState` interface already defined the type : https://github.com/kuzzleio/kuzzle-admin-console/blob/e2ff7711582abc27a7f3d0160f2ae5aacc19f96c/src/vuex/modules/index/types.ts#L84

src/components/Data/Leftnav/IndexBranch.vue
```diff
@@ -95,10 +105,16 @@ export default {
...
+      showMoreCollections: false,
+      collectionsFetched: false,
+      isLoading: false
     }
   },
   computed: {
+    ...mapGetters('index', ['loadingCollections']),
+    // isLoading() {
+    //   return this.loadingCollections(this.index.name)
+    // },
```
xbill82, line 117:
```suggestion

```

src/components/Data/Collections/Update.vue
```diff
@@ -69,12 +69,11 @@ export default {
   methods: {
     async update(payload) {
       this.error = ''
-
+      console.log(payload)
```
xbill82:
```suggestion

```

PR #834, opened 2020-10-29
add v4 banner

### LINKED ISSUE #803: Add banner to show the existence of the v4
Let's add a banner on top of the v3 to tell the users they can try the v4 on next-console.kuzzle.io

berthieresteban:
## What does this PR do ?
resolve #803 
Add a banner telling users they should try the new admin console v4

  - Step 1 : kourou instance:spawn
  - Step 2 : check signup, login and home pages

### Other changes
/

### Boyscout
/

Capture du 2020-10-29 10-33-26
Capture du 2020-10-29 10-32-07
Capture du 2020-10-29 10-31-58


xbill82 (approved):
Cool! Can we add some horizontal padding to the banners in the login page and the signup page?

berthieresteban:
Did it but I had to add a break line since the full text doesn't fit the line and I can't set col 12 because it would be ugly not being aligned with text inputs
Capture du 2020-10-29 15-56-00

Njuelle:
It is also necessary to specify that it is a beta version.

Njuelle (approved):

PR #849, opened 2020-11-17
Revoke anonymous role right

morgandruesne:
## What does this PR do ?
Added a button to revoke the rights of anonymous roles in the roles page


Njuelle (changes requested):
IMHO before showing this button, we need to check if there is an administrator account created, and the current user is not anonymous (there is a getters `adminAlreadyExists` in the `auth` store)

src/components/Security/Roles/Page.vue
```diff
@@ -26,14 +34,21 @@
...
         </p>
       </b-card>
     </role-list>
+    <b-modal
+      id="modal-revokeAnonymous"
+      title="Revoke anonymous rights"
+      @ok="revokeAnonymous"
+    >
+      <p class="my-4">
+        Do you want to revoke the rights of the annonymous role ?
```
xbill82, line 49: Let's add a text that explains the consequences, something like

> The anonymous users will only be able to perform some basic authentication actions, like logging-in, see their rights and see their user ID. You will still be able to add more rights if needed.

src/components/Security/Roles/Page.vue
```diff
@@ -5,6 +5,14 @@
         <headline>Roles</headline>
       </b-col>
       <b-col class="text-right mt-3">
+        <b-button
+          v-if="displayRevokeAnonymous"
```
xbill82, line 9:
```suggestion
          :disabled="!displayRevokeAnonymous"
          :title="displayRevokeAnonymous ? 'Reduce anonymous rights to the minimum' : 'You cannot revoke anonymous rights either because you don\'t have the permissions to do it or because there is no administrator user yet in the system.'
```

xbill82 (changes requested):

xbill82 (changes requested):
... aaaaand tests are missing 😬

src/components/Security/Roles/Page.vue
```diff
@@ -50,8 +65,65 @@ export default {
...
+        }
+      });
+      this.$router.go(this.$router.currentRoute)
+    },
+  },
   computed: {
-    ...mapGetters('auth', ['canSearchRole', 'canCreateRole'])
+    ...mapGetters('kuzzle', ['$kuzzle']),
+    ...mapGetters('auth', ['canSearchRole', 'canCreateRole']),
+    displayRevokeAnonymous () {
```
Njuelle, line 122: you also need to check if the current user is not Anonymous

src/components/Security/Roles/Page.vue
```diff
@@ -50,7 +64,59 @@ export default {
     RoleList,
     Headline
   },
+  methods: {
+    revokeAnonymous () {
```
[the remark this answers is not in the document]
Njuelle: still no error if something goes wrong ;)

Njuelle (changes requested):

src/components/Security/Roles/Page.vue
```diff
@@ -50,8 +65,65 @@ export default {
...
+          },
+          server: {
+            actions: {
+              publicApi: true,
+              openapi: true
+            }
+          }
+        }
+      });
+      this.$kuzzle.security.updateRole('default', {
```
Aschen, line 93: This promise must be awaited or a catch handler must be attached to it

src/components/Security/Roles/Page.vue
```diff
@@ -50,8 +65,65 @@ export default {
     RoleList,
     Headline
   },
+  methods: {
+    revokeAnonymous () {
+      this.$kuzzle.security.updateRole('anonymous', {
```
Aschen, line 70: This promise must be awaited or a catch handler must be attached to it

Aschen (changes requested):

xbill82 (approved):

PR #864, opened 2020-12-28
838 fix displayed id

### LINKED ISSUE #838: ID displayed isn't good when document include a field "id"
## Context

Hi,
I don't know if it's a feature or a bug but I spent a lot of time to understand what happens.

It's simple, when documents include a field "id", the console use this field as identifier in view line and not real document id.
In my usage I generate a specific ID for documents to get them quickly and "id" field in my document store other information.

So, as I said, I don't know if it's a bug or a feature, but I lost time to find solution and it's when I used directly API that I saw that ID was ok after my imports. 

## Expected Behavior
See view line of documents with good document _id

## Current Behavior

We saw field "id" in document and not reald document _id

## Possible Solution

To "fix" this I rename field id and reimport documents

## Steps to Reproduce

1. Import document this a field "id"
2. Show them in Console next generation

## Context (Environment)

Kuzzle version: 2.6.0
Node.js version: v12.13.1
Admin Console Version: 4-dev (http://next-console.kuzzle.io/)

berthieresteban:
## What does this PR do ?
fix #838 

  - Step 1 : create a collection with field `id` in mapping
  - Step 2 : create a document with a value in `id` field
  - Step 3 : check on document views (list/column/map...), the _id is now  displayed
  ...

### Other changes
disable map view when mapping have no `geo_point` field
disable chart view when mapping have no `integer` field

### Boyscout
catch json editor error (browser console was spammed when creating/editing documents)

### DIFF
--- a/src/components/Data/Documents/Common/CreateOrUpdate.vue
+++ b/src/components/Data/Documents/Common/CreateOrUpdate.vue
@@ -202,7 +202,7 @@ export default {
         parsed = JSON.parse(val)
         this.$emit('document-change', parsed)
       } catch (error) {
-        // invalid json
+        // Fail silently
       }
     },
     onFormChange() {
--- a/src/components/Data/Documents/Common/CreateOrUpdate.vue
+++ b/src/components/Data/Documents/Common/CreateOrUpdate.vue
@@ -93,6 +93,7 @@
             data-cy="DocumentCreate-btn"
             variant="primary"
             class="ml-2"
+            :disabled="submitting || !isDocumentValid"
             @click="submit"
           >
             <i class="fa fa-plus-circle left" />
@@ -196,7 +197,13 @@ export default {
   methods: {
     onJsonChange(val) {
       this.rawDocument = val
-      this.$emit('document-change', JSON.parse(val))
+      let parsed = {}
+      try {
+        parsed = JSON.parse(val)
+        this.$emit('document-change', parsed)
+      } catch (error) {
+        // invalid json
+      }
     },
     onFormChange() {
       this.rawDocument = JSON.stringify(this.document, null, 2)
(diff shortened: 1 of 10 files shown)


berthieresteban:
Draft -> I forgot to write some tests

src/components/Data/Documents/Common/CreateOrUpdate.vue
```diff
@@ -196,7 +197,13 @@ export default {
   methods: {
     onJsonChange(val) {
       this.rawDocument = val
-      this.$emit('document-change', JSON.parse(val))
+      let parsed = {}
+      try {
+        parsed = JSON.parse(val)
+        this.$emit('document-change', parsed)
+      } catch (error) {
+        // invalid json
```
alexandrebouthinon, line 205: Maybe we should do something in case of error. At least log it into the JS console

alexandrebouthinon (approved):

src/components/Data/Documents/Common/CreateOrUpdate.vue
berthieresteban: That method is called every time the text in json editor changed, I'm not sure we want to print an error in console when user is writing something and haven't finished so the json can't be parsed :/
For example, just write `{"test":"test"}` will print at least 7 error in console

berthieresteban: That method is called every time the text in json editor changed, I'm not sure we want to print an error in console when user is writing something and haven't finished so the json can't be parsed :/
For example, just write `{"test":"test"}` will print at least 7 error in console

Leodau, line 205: If we do not want to log or do anything with this error, then add an explicit comment.
```suggestion
        // Fail silently
```

Leodau (approved):
This bug was killing me :muscle:, clean fix.
I will still wait for someone from the AC team to merge, tho.

xbill82 (approved):
🎉

PR #883, opened 2021-01-15
Check if user is authenticated before lazy loading sequence

Njuelle:
## What does this PR do ?
Check if user is authenticated before lazy loading sequence, using auth `isAuthenticaded` store getter

  - Step 1 : log in to an environment then log out
  - Step 2 : log in to another environment
  - Step 3 : try to switch to the first environment
  - Step 4 : check that there are no errors and no error toast displayed

### Other changes
- force Kuzzle-SDK to version 7.4.7 to avoid TS errors until the sdk patch is released

### DIFF
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -158,8 +158,12 @@ export default {
     handleDataNotFound() {
       this.dataNotFound = true
     },
-    async fetchAllTheThings() {
+    // @todo : handle lazy loading sequence only on the authenticated routes
+    async lazyLoadingSequence() {
       if (!this.isAuthenticated) {
+        this.$log.warn(
+          'Lazy loading sequence started with a non-authenticated user.'
+        )
         return
       }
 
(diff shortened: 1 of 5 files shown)


rolljee (approved):

src/components/Data/Layout.vue
```diff
@@ -158,6 +159,10 @@ export default {
       this.dataNotFound = true
     },
     async fetchAllTheThings() {
+      if (!this.isAuthenticated) {
+        return
```
xbill82, line 163: 🤔 I feel the day the AC will fall into this case outside a login or bootstrap sequence. The page will be empty, nothing will happen, no error in the console... Just crickets.
I suggest that, for the time being, we just add a warning in the console and when the moment comes, we handle this properly.

xbill82 (approved):
I was wondering if we could test this case. Not sure we can.

src/components/Data/Layout.vue
Njuelle: I totally agree!

Njuelle: Also, I found a way to test this case :)

PR #891, opened 2021-01-23
Expose an instantiated and connected SDK in the window object

Aschen:
## What does this PR do ?

Expose an instantiated, connected an authenticated SDK in the `window` object.

Now you can open the browser console and play with the SDK directly
image


Leodau (approved):
Kindda like this :ok_man:

Njuelle (approved):

PR #894, opened 2021-01-27
Fix Json-editor multiple instance

### LINKED ISSUE #893: [bug] Cannot use form-view with multiple instances of JsonFormInput
Cannot use form-view with multiple instances of JsonFormInput

Njuelle:
## What does this PR do ?
Fix #893  Json-editor multiple instance

  - Step 1 : Create a collection with at least 2 nested properties
```
{
  "bar": {
    "properties": {
      "barfoo": {
        "type": "keyword"
      },
      "bazfoo": {
        "type": "keyword"
      }
    }
  },
  "foo": {
    "properties": {
      "foobar": {
        "type": "keyword"
      },
      "foobaz": {
        "type": "keyword"
      }
    }
  }
}
```
  - Step 2 : try create and edit a document with the form-view enabled


src/components/Common/JsonEditor.vue
```diff
@@ -26,70 +27,77 @@
...
       type: String,
       default: ''
     },
     readonly: Boolean,
-    id: String,
     height: { type: Number, default: 250 }
   },
+  data() {
+    return {
+      editor: null
```
xbill82, line 49: What is the purpose of putting the `editor` variable in the component state?

xbill82 (commented):
It would be nice if we had the same code-formatting rules.

src/components/Common/JsonEditor.vue
Njuelle: it is necessary to have the `editor` variable scoped to the component to be able to use several instances

jenow (approved):

src/components/Common/JsonEditor.vue
xbill82: why?

test/e2e/cypress/integration/single-backend/formView.spec.js
```diff
@@ -38,6 +44,10 @@ describe('Form view', function() {
           phone: 'Blackberry',
           car: 'Laguna'
         },
+        skill: {
+          name: 'managment',
+          level: '1'
+        },
         job: 'Always asking Esteban to do his job',
         employeeOfTheMonthSince: '1996-07-10'
       }
```
berthieresteban, line 53: :heart:

src/components/Common/JsonEditor.vue
berthieresteban: If we let the `editor` in global, if we add multiple json-editors there will still be only one `editor` variable wich will be overrided each time we instanciate a new json-editor.
So call setContent method of a json-editor component will always set the content on the last instanciated editor no matter wich json-editor's method we call

xbill82: I see. The `editor` variable is scoped to the JSONEditor.vue module, which is a singleton, in the CommonJS modules model. The module exports a VueJS Component config, which is passed to the `Vue` constructor each time this component is instantiated. But this config has a reference to the `editor` variable, which is actually overwritten, as you rightly say, each time the component is instantiated. By putting the variable inside the `data` structure, you make it local to the component instead, which is obviously not a singleton. This is a nice thing to keep in mind for future developments.

I still don't like the idea of putting a complex object (i.e. one that has attributes and functions) inside the `data` structure, because it is meant to keep track of the _state_ of the component, yet this object is just logic to make the editor work. We should find a way to replicate this reference each time the component is created, and avoid the reference to be overwritten.

I am not really hurried to ship this fix, but if you are, we can keep this enhancement for later. What do you think?

berthieresteban: After merge this PR I'll create an issue concerning the refacto of the json-editor and we'll probably need a workshop

berthieresteban (approved):

PR #920, opened 2021-06-04
906 bad translation

### LINKED ISSUE #906: Bad translation
`Not Authentified` https://github.com/kuzzleio/kuzzle-admin-console/blob/master/src/components/Common/MainMenu.vue#L79

berthieresteban:
fix #906

### DIFF
--- a/src/components/Common/MainMenu.vue
+++ b/src/components/Common/MainMenu.vue
@@ -96,7 +96,7 @@ export default {
     },
     currentUserName() {
       if (!this.user) {
-        return 'Not authentified'
+        return 'Not authenticated'
       }
       if (this.user.id === -1) {
         return 'Anonymous'


Aschen (approved):

src/components/Common/MainMenu.vue
```diff
@@ -96,7 +96,7 @@ export default {
     },
     currentUserName() {
       if (!this.user) {
-        return 'Not authentified'
+        return 'Not authenticated'
```
Aschen, line 99: Why it's not `Anonymous` instead ?

berthieresteban: It is different
```
      if (!this.user) {
        return 'Not authenticated'
      }
      if (this.user.id === -1) {
        return 'Anonymous'
      }
```

Aschen: Actually for Kuzzle it's not since when a request is made without authentication Kuzzle automatically assign the anonymous user.
But maybe in the AC it's different so I will let you judge which way is better on this one

xbill82: Actually, the AC has a slightly different state machine:
* Not authenticated (`currentUser === null`) means the end user has not yet decided whether to start the current session as an authenticated user, or the Anonymous user 👉 we display the login dialog.
* Anonymous (`currentUser === 'Anonymous'`) means the user has decided to start the session as the Anonymous user, thus uses the app with the rights attached to this user.
* Another user (`currentUser === 'luca'`) means the user has decided to login as an authenticated user, which ends up being the same case as the previous one.

xbill82 (approved):

Leodau (approved):

PR #921, opened 2021-06-04
fix perpage change bug

### LINKED ISSUE #917: Table doesn't display all of the items when changing numbers of items per page
## Expected Behavior
List all the items in the table.

## Current Behavior
When we are on a page that is not the first and we change the number of items to display per page it's not displaying all the items but only those on this page and on the pages after.

Capture d’écran du 2021-04-16 12-08-59

## Steps to Reproduce
1. Go on the last page of a table
2. Change the numbers of items per page

berthieresteban:
fix #917

### DIFF
--- a/src/components/Common/CommonList.vue
+++ b/src/components/Common/CommonList.vue
@@ -113,7 +113,7 @@ export default {
       return parseInt(this.currentFilter.from) || 0
     },
     paginationSize() {
-      return parseInt(this.currentFilter.size) || 10
+      return parseInt(this.currentFilter.size) || 25
     }
   },
   watch: {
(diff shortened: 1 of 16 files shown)


Aschen (commented):
I know it's not really related but could we increase the default number of documents per page to 25 instead of 10? 
10 is really not a lot

berthieresteban:
Yes we could, should we talk about this or I can directly add that in this PR ?

xbill82 (approved):

xbill82:
@berthieresteban let's onboard the new default value in this PR

Leodau (approved):

PR #922, opened 2021-06-04
fix initial filter displayed value issue

### LINKED ISSUE #908: [bug] Correctly update the quick-search input
## Expected Behavior
Correctly have quick-search in sync with the interface.

## Current Behavior
When changing pages **after** doing a search and coming back, the query is saved and applied BUT the search input is empty.
(Ghost query effect)

_It seems like the search queries are correctly saved and re-applied on page refresh, but the search input control is not in sync._

## Steps to Reproduce
1. Go to the Security/User tab.
2. Create a new user (_kuid_ test)
3. Back at the User tab, execute a quick-search like: `alyx` _(You should now have an empty result)_
4. Click and go to another tab (_ex. Profiles)_
5. Click and come back to the User tab. _(You should see that your **input** is empty, but your query is still being applied)_

Or,
1. Make any quick-search query
2. Reload the page

## Context (Environment)
Kuzzle version: 2.9.2
Admin Console version: 4

berthieresteban:
fix #908

### DIFF
--- a/src/components/Security/Users/List.vue
+++ b/src/components/Security/Users/List.vue
@@ -3,6 +3,7 @@
     <slot v-if="isCollectionEmpty && !loading" name="emptySet" />
     <template v-else>
       <filters
+        v-if="!initialLoading"
         class="mb-3"
         :available-operands="searchFilterOperands"
         :current-filter="currentFilter"
@@ -194,6 +195,7 @@ export default {
       deleteModalIsLoading: false,
       documents: [],
       loading: true,
+      initialLoading: true,
       searchFilterOperands: filterManager.searchFilterOperands,
       selectedDocuments: [],
       totalDocuments: 0,
@@ -397,6 +399,7 @@ export default {
     }
   },
   mounted() {
+    this.initialLoading = true;
     this.currentFilter = filterManager.load(
       this.index,
       this.collection,
@@ -408,6 +411,7 @@ export default {
       this.index,
       this.collection
     )
+    this.initialLoading = false;
   },
   watch: {
     $route: {


xbill82 (changes requested):
I don't agree with this fix. This is symptomatic that the `Filters` component is not correctly reactive against the `currentFilter` state variable, which means there is something bad going on under the hood. What do you think?

berthieresteban:
I agree with you
It was a quickfix to resolve the bug, I just didn't have more time to spend
Also, I'm going to look a little bit more because with the history system it is possible to load a filter so the problem may not come from the Filters component but from the way it is used in this view

xbill82:
Thanks @berthieresteban ! This is why I'm always afraid of two-way bindings...

xbill82:
Yo, this implementation is valid, but... May I ask you to try to just use (in the `b-input`) 
```
:value="value"
@input="$emit('input', $event)
```
So that the component becomes stateless?

Just to try if it works...

berthieresteban:
Could we discuss again about the auto-focus feature in this component ?
I really don't like the implementation...
I don't even know there was that feature in the admin console, is it still used actually ? 
The `autofocus` prop from bootstrap-vue could be enough ?

Aschen:
@xbill82 @Njuelle  updates on this?

xbill82:
@Aschen the auto-focus input is a feature you added, so we wanted to discuss it with you. Can we call a quick meeting or do you prefer to discuss it here asynchronously?

Aschen:
We can remove it :+1:

Njuelle (approved):

PR #925, opened 2021-06-04
fix strategies displayed error

### LINKED ISSUE #873: User body display strategy without credentials
## Expected Behavior
In security user list, the users items display all available strategy in json, even if user has no credentials for that strategy.

## Current Behavior
Empty strategies should not be displayed

## Possible Solution
remove [line 161 of kuzzleWrapper-v1.ts](https://github.com/kuzzleio/kuzzle-admin-console/blob/f4e54ea14a3dbf3835e92ea9e8aa0d2d41209c15/src/services/kuzzleWrapper-v1.ts#L161)

berthieresteban:
fix #873


xbill82 (approved):

Njuelle (approved):

Njuelle:
it seems that the tests must be fixed ;)

berthieresteban:
This test should pass, the `local` strategy is still displayed as a `key` in the `strategies` object and the `value` is empty so it has the class `json-formatter-empty`
I just removed the `Username` `Password` keys...
Also, the test pass locally

berthieresteban:
I did remove the `local` property in `credentials` since there is nothing in it...
Test should be fixed

PR #927, opened 2021-06-16
Display local strategy usernames if found

### LINKED ISSUE #832: Improve the user list UI
## Feature description

We should display the username from the local strategy because UUID only are not human-readable.

Adding a small action to copy the ID could be nice as well.

Leodau:
## What does this PR do ?
As the title suggests, on the **Security:Users** page, display the user local strategy `username` as the list item title **if one is found**.

Clone, Serve and [head to the users page](http://localhost:8080/#/security/users) to see that user items can now be easily identified.

image

Closes #832

### DIFF
--- a/src/components/Security/Users/UserItem.vue
+++ b/src/components/Security/Users/UserItem.vue
@@ -1,6 +1,6 @@
 <template>
-  <div>
-    <b-container fluid data-cy="UserItem" class="UserItem">
+  <div data-cy="UserItem" class="UserItem">
+    <b-container fluid class="UserItem-container">
       <div class="UserItem-toggle-select mr-3">
         <i
           aria-role="button"
@@ -178,9 +178,11 @@ export default {
 
 <style lang="scss" rel="stylesheet/scss" scoped>
 .UserItem {
-  display: flex;
-  flex-direction: row;
-  align-items: center;
+  &-container {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+  }
 
   &-toggle-select {
     display: flex;
--- a/src/components/Security/Users/UserItem.vue
+++ b/src/components/Security/Users/UserItem.vue
@@ -25,7 +25,7 @@
         <a
           class="d-inline-block align-middle code pointer mr-2"
           @click="toggleCollapse"
-          >{{ document.id }}</a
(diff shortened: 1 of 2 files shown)


Aschen (commented):
It's confusing to me because it means that the user view is the only one that does not display document IDs in the AC.

Also sometime this view will displays IDs and sometime local strategy usernames.

I think we should rather display the username alongside the ID if it exists

xbill82 (approved):

Aschen (changes requested):
It's confusing to me because it means that the user view is the only one that does not display document IDs in the AC.

Also sometime this view will displays IDs and sometime local strategy usernames.

I think we should rather display the username alongside the ID if it exists

xbill82:
Ah, @Aschen good idea!

Leodau:
> Ah, @Aschen good idea!

Makes sense to me, the thing is… we don't have tons of space to begin with :grimacing: 
image

Aschen:
Maybe we could have a switch between "Display kuid" and "Display auth local" ?

xbill82:
@Leodau I'd prefer a little design effort in order to show them both. We can play with the shades of grey (username matte black and ID grey just below and tiny, for example). And we can ask advice to @jsatge !

berthieresteban:
Is the profiles information absolutely necessary there ?
There is my analysis:
  - The username information is more important than the profiles information because in that list we want to find quickly the user we are searching for, also to see user content we need to expand the div where we can find the profiles information
  - Actually, the kuid is always displayed entirely but the profiles section can grow up based on the number of profiles so the kuid can be displayed on two lines with a 150px width and the profiles badges may not be all shown
  - Most of the time, the kuid and the username can be displayed one next to the other if we don't display the profile information there
  
IMO: 
We could display the kuid | username (if exists) without the profiles
And add an additionnal and easy to use filter above the list to search on a profile (like in the profile section to search on the roles)
That would allow us to always display user on a single line and that would make easier to search users by profiles

Njuelle:
Honestly, having the profiles displayed in this list is really useful, I suggest something like this :
image

Njuelle:
final result
image

jenow (approved):

Aschen (approved):

PR #930, opened 2021-06-17
Fix/Increase the search profile count to improve user creation drop-down

### LINKED ISSUE #892: [bug] the search profile action limits the size of the results to 100
the `performSearchProfiles` action on the `KuzzleWrapper` service limit the size of the results to 100

https://github.com/kuzzleio/kuzzle-admin-console/blob/4-dev/src/services/kuzzleWrapper-v1.ts#L206

Leodau:
## What does this PR do ?

As the title suggests, increased the number of requested profiles, so the user creation drop-down correctly show at least 1k profiles.

Closes #892

### DIFF
--- a/src/services/kuzzleWrapper-v1.ts
+++ b/src/services/kuzzleWrapper-v1.ts
@@ -205,7 +205,7 @@ export class KuzzleWrapperV1 {
   async performSearchProfiles(filters = {}, pagination = {}) {
     const result = await this.kuzzle.security.searchProfiles(
       { ...filters },
-      { size: 100, ...pagination }
+      { size: 1000, ...pagination }
     )
 
     const profiles = result.hits.map(document => {


xbill82 (approved):

berthieresteban:
IMHO we should add a scrolled pagination to the select, I know it's not possible as a quickfix so I would prefer adding a parameter like `this.wrapper.performSearchProfiles({}, { size: 1000 })` in  [userProfileList](https://github.com/kuzzleio/kuzzle-admin-console/blob/1edcd347425b048d91895c1ade316bf9fc7c2093/src/components/Security/Users/Steps/UserProfileList.vue#L95) to avoid side effect by changing the default behavior of the `performSearchProfiles`

jenow (approved):

PR #932, opened 2021-07-01
Add import/export API actions

berthieresteban:
## What does this PR do ?
Add import/export features to API Actions view

Buttons import/export to left corner in sidebar
export:
  - open modal with table listing all the saved queries of all env (in the localStorage)
  - allow to check one or more queries
  - export button modal-ok

import: 
  - open modal with file input (file explorer/drag n drop)
  - multifile import supported
  - list of files to import (size and remove button)
  - list of queries to import 
  - allow to edit queries name before import (inputstate (red/green) = name already used)

Tests added:
  - Export a query in file and check file content
  - Import a query with file 
  - Change a query name while importing

  - Step 1 : create and save some queries
  - Step 2 : click the export button and check the file
  - Step 3 : delete your previous saved queries
  - Step 4 : click the import button and upload the previous downloaded file

### Other changes
/

### Boyscout
/

Screenshot from 2021-07-01 17-50-55

Screenshot from 2021-07-01 17-59-37

Screenshot from 2021-07-01 17-59-17

### DIFF
--- a/src/components/ApiAction.vue
+++ b/src/components/ApiAction.vue
@@ -38,87 +38,80 @@
       <b-container fluid class="h-100">
         <b-row class="h-100">
           <b-col cols="12" v-if="loading"> </b-col>
-          <b-col cols="12" v-else>
-            <b-row class="h-100">
-              <b-col cols="12">
-                <b-card no-body class="px-0 h-100">
-                  <b-tabs card content-class="px-0 mt-3 tabsHeight">
-                    <b-tab
-                      v-for="(tabContent, tabIdx) of tabs"
-                      :key="`query-${tabIdx}-${tabContent.name}`"
-                      :active="currentTabIdx === tabIdx"
-                      @click.prevent=""
-                      class="px-2 py-0"
-                      title-link-class="px-3 py-0 titleItem"
+          <b-col cols="12" class="h-100" v-else>
+            <b-card no-body class="px-0 h-100">
+              <b-tabs card content-class="px-0 mt-3 tabsHeight">
+                <b-tab
+                  v-for="(tabContent, tabIdx) of tabs"
+                  :key="`query-${tabIdx}-${tabContent.name}`"
+                  :active="currentTabIdx === tabIdx"
+                  @click.prevent=""
+                  class="px-2 py-0"
+                  title-link-class="px-3 py-0 titleItem"
+                >
+                  <template #title>
+                    <b-row
+                      align-v="center"
+                      class="tabTitle"
+                      v-b-tooltip.hover
+                      :data-cy="`api-actions-tab-${tabIdx}`"
+                      :title="tabContent.name"
                     >
-                      <template #title>
-                        <b-row
-                          align-v="center"
-                          class="tabTitle"
-                          v-b-tooltip.hover
-                          :data-cy="`api-actions-tab-${tabIdx}`"
-                          :title="tabContent.name"
-                        >
-                          <b-col
-                            cols="9"
-                            class="text-left py-3 pointer"
-                            @click="setCurrentTab(tabIdx)"
-                          >
-                            <span>
-                              {{ formatTabName(tabContent) }}
-                            </span>
-                          </b-col>
-                          <b-col cols="3" class="py-3">
-                            <i
-                              class="fas fa-times pointer"
-                              @click="closeTab(tabIdx)"
-                            />
-                          </b-col>
-                        </b-row>
-                      </template>
-                      <QueryCard
-                        class="px-0"
-                        :query="tabContent.query"
-                        :tabIdx="tabIdx"
-                        :api="api"
-                        :openapi="openapi"
-                        :response="tabContent.response"
-                        @saveQuery="saveQuery"
-                        @queryChanged="onQueryChanged"
-                        @performQuery="performQuery"
-                      />
-                    </b-tab>
-                    <template #tabs-end>
-                      <b-row
-                        align-v="center"
-                        class="px-3"
-                        @click.prevent="addNewTab"
+                      <b-col
+                        cols="9"
+                        class="text-left py-3 pointer"
+                        @click="setCurrentTab(tabIdx)"
                       >
-                        <b-col cols="12" class="text-center my-3 pointer">
-                          <b data-cy="api-actions-tab-plus">+</b>
-                        </b-col>
-                      </b-row>
-                    </template>
-                    <template #empty>
-                      <b-row align-v="center" align-h="center" class="h-100">
-                        <b-col cols="4">
-                          <b-card title="No API action opened.">
-                            <b-card-text>
-                              <p>
-                                You can open a saved action in the left menu or
-                                <b-button variant="primary" @click="addNewTab"
-                                  >create one</b-button
-                                >
-                              </p>
-                            </b-card-text>
-                          </b-card>
-                        </b-col>
-                      </b-row>
-                    </template>
-                  </b-tabs>
-                </b-card>
-              </b-col>
-            </b-row>
+                        <span>
+                          {{ formatTabName(tabContent) }}
+                        </span>
+                      </b-col>
+                      <b-col cols="3" class="py-3">
+                        <i
+                          class="fas fa-times pointer"
+                          @click="closeTab(tabIdx)"
+                        />
+                      </b-col>
+                    </b-row>
+                  </template>
+                  <QueryCard
+                    class="px-0"
+                    :query="tabContent.query"
+                    :tabIdx="tabIdx"
+                    :api="api"
+                    :openapi="openapi"
+                    :response="tabContent.response"
+                    @saveQuery="saveQuery"
+                    @queryChanged="onQueryChanged"
+                    @performQuery="performQuery"
+                  />
+                </b-tab>
+                <template #tabs-end>
+                  <b
+                    @click.prevent="addNewTab"
+                    class="px-3 my-3 text-center pointer"
+                    data-cy="api-actions-tab-plus"
+                    >+</b
+                  >
+                </template>
+                <template #empty>
+                  <b-row align-v="center" align-h="center" class="h-100">
+                    <b-col cols="4">
+                      <b-card title="No API action opened.">
+                        <b-card-text>
+                          <p>
+                            You can open a saved action in the left menu or
+                            <b-button variant="primary" @click="addNewTab"
+                              >create one</b-button
+                            >
+                          </p>
+                        </b-card-text>
+                      </b-card>
+                    </b-col>
+                  </b-row>
+                </template>
+              </b-tabs>
+            </b-card>
           </b-col>
         </b-row>
       </b-container>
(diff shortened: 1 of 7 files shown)


src/components/ApiAction/ExportActionsModal.vue
```diff
@@ -0,0 +1,115 @@
...
+      apiActions: [],
+      selectedActionIds: []
+    }
+  },
+  computed: {
+    exportURL() {
+      const selectedActions = this.apiActions.filter(action =>
+        this.selectedActionIds.includes(action.id)
+      )
+      selectedActions.map(action => {
```
Aschen, line 56: Why using `.map` only to mutate items ? You should use `forEach`

src/components/ApiAction/ExportActionsModal.vue
```diff
@@ -0,0 +1,115 @@
...
+        delete action.id
+        delete action.idx
+        delete action.env
+      })
+      const data = JSON.stringify(selectedActions)
```
Aschen, line 61: Can you prettify the JSON?

src/components/ApiAction/ExportActionsModal.vue
```diff
@@ -0,0 +1,115 @@
...
+
+      const blob = new Blob([JSON.stringify(data)], {
```
Aschen, line 63: Why using stringify 2 times?

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,192 @@
...
+          variant: 'danger',
+          toaster: 'b-toaster-bottom-right',
+          appendToast: true,
+          dismissible: true,
+          noAutoHide: true
+        })
+        return true
+      }
+    },
+    async readFile(file) {
```
Aschen, line 149: Useless async

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,192 @@
...
+          }
+        }
+        reader.onerror = evt => {
+          reject()
+        }
+      })
+    },
+    async fileAdded() {
+      const isInvalidFile = await this.isInvalidFile(this.file)
+      if (!isInvalidFile) {
```
Aschen, line 167: Instead of having `if (! isInvalidFile)` you could just have `if (isValidFile)`. Usually simple condition are easier to read

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,192 @@
...
+        this.files.push(this.file)
+      }
+      this.file = null
+    },
+    async filesChange() {
+      const fileContent = {}
+      for (const file of this.files) {
+        fileContent[file.name] = await this.readFile(file)
```
Aschen, line 175: Those calls to `readFile` should be done in //

Aschen (changes requested):

src/components/ApiAction/QueryList.vue
```diff
@@ -18,33 +15,35 @@
...
-                    </b-col>
-                  </b-row>
-                </b-list-group-item>
-              </b-list-group>
+              <b-card no-body header="Saved API Actions">
+                <b-list-group class="leftNav-container" ref="leftNav-container">
+                  <b-list-group-item
+                    v-for="query of paginatedQueries"
+                    class="px-3 py-0"
+                    :key="`saved-query-${query.idx}`"
```
xbill82, line 23:
```suggestion
                    :key="`saved-query-${query.name}`"
```
Using the index of the array element doesn't help the VDom library to resolve which element should be re-rendered and how, since indexes change when the array is mutated. Use instead more business-related properties, like the name or some property that really differentiates one element from the others.

xbill82 (changes requested):
Very good job! 
Just a few things to change 😄

src/components/ApiAction.vue
```diff
@@ -12,74 +12,113 @@
...
+                      @click.prevent=""
+                      class="px-2 py-0"
+                      title-link-class="px-3 py-0 titleItem"
                     >
-                      <b-col
-                        cols="9"
-                        class=" text-left py-3 pointer"
-                        @click="setCurrentTab(tabIdx)"
+                      <template #title>
+                        <b-row
```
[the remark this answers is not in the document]
berthieresteban: For this one it is just to use the grid system to specify col size between text and icon
I really dont feel like it's weird to use a b-row in a component that especially look like a row

src/components/ApiAction.vue
```diff
@@ -12,74 +12,113 @@
...
-                  title-link-class="px-3 py-0 titleItem"
-                >
-                  <template #title>
-                    <b-row
-                      align-v="center"
-                      class="tabTitle"
-                      v-b-tooltip.hover
-                      :data-cy="`api-actions-tab-${tabIdx}`"
-                      :title="tabContent.name"
+            <b-row class="h-100">
```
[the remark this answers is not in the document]
berthieresteban: Just because I think this is how it should be used, it make no sense to me to put col in col or row in row directly but I will change that

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,192 @@
...
+      <b-button
+        @click="handleImport"
+        :disabled="validNames || !Boolean(files.length)"
+        variant="primary"
+        data-cy="modal-button-import-action"
+        >Import
+      </b-button>
+    </template>
+    <b-container>
+      <b-row>
```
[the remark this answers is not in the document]
berthieresteban: Div doesn't take the entire width in a container, div doesn't have a default harmonious padding/margin, use div involves to add classes like flex for example ? 
Personaly I prefer to use rows and col because I'm really familiar with the grid system... I can change it anyway

src/components/ApiAction/QueryList.vue
berthieresteban: Actually, with a localStorage manipulation we could have two queries with the same name, I used idx to avoid key duplicate and so errors

src/components/ApiAction.vue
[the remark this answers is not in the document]
berthieresteban: Ok I see there was an extra row you are right !

src/components/ApiAction/QueryList.vue
xbill82: That's the difficult part about item lists with frameworks based on vDom. You should really find a way to create a string that uniquely identifies each item. And if you have items that are not unique (meaning they have the same exact content), then it's not a problem for the vDom, because it means the item doesn't need to be re-rendered.

berthieresteban: In the app we don't allow duplicates names for API Actions, the only way to there could be a duplicate is if an user manipulate directly the local storage
Should we handle this too ? What should be the behavior in that case ?
In case of a duplicate which API Action should we hide ? If we do that the user won't be able to save it with another name
At least we could just check if there are duplicates, hide them and display a warning toast to tell the user check the localStorage ?

xbill82: no no, I don't think it's a good idea to allow to save duplicate actions, I was only talking about it in general terms

xbill82:
Er... What happened to the Netlify previews? 👀

src/components/ApiAction/QueryList.vue
berthieresteban: I wasn't talking about allowing save duplicate but how to handle the case when a user add duplicate directly in localStorage

Njuelle (approved):

src/components/ApiAction/QueryList.vue
xbill82: oh, ok

xbill82:
I need to dive back into the code but, looking at the PR description, I find it weird to be able to export the API actions for all the environments, given that one cannot see other environments' actions in the list on the left sidebar. To me
* Either API actions are cross-environment (i.e. all the actions are always visible no matter to which Kuzzle the AC is connected to) and therefore exportable
* Actions are separated by environment, so one can only export the actions of the current environment (i.e. the current connection).

What do you think @berthieresteban @Aschen @Njuelle ?

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,199 @@
...
+                  size="sm"
+                  :data-cy="`imported-action-name-editor-${data.index}`"
+                  :state="!savedActionNames.includes(data.item.name)"
+                  v-model="data.item.name"
+                >
+                </b-form-input>
+                <b-form-invalid-feedback
+                  :id="`name-input-${data.index}-feedback`"
+                >
+                  You already have a saved action with that name.
```
xbill82, line 64: Note for self 🤣 
```suggestion
                  An action with this name already exists.
```

src/components/ApiAction/ImportActionsModal.vue
```diff
@@ -0,0 +1,199 @@
...
+        { key: 'name', label: 'Name' },
+        { key: 'query.controller', label: 'Controller' },
+        { key: 'query.action', label: 'Action' }
+      ]
+    }
+  },
+  computed: {
+    validNames() {
+      return this.actions.some(action =>
+        this.savedActionNames.includes(action.name)
```
xbill82, line 99: This looks weird

src/components/ApiAction/QueryList.vue
xbill82: I think it's a good idea to add a duplicate-management mechanism. I'll try to figure something out to prevent duplicates to land on the action list.

xbill82: In the meanwhile, I'd prefer to use the action name as the v-for key, which looks safe enough.

berthieresteban:
> I need to dive back into the code but, looking at the PR description, I find it weird to be able to export the API actions for all the environments, given that one cannot see other environments' actions in the list on the left sidebar. To me
>
> * Either API actions are cross-environment (i.e. all the actions are always visible no matter to which Kuzzle the AC is connected to) and therefore exportable
> * Actions are separated by environment, so one can only export the actions of the current environment (i.e. the current connection).

Your first point is convenient, the second is structured

IMHO Both have pro and cons
 * First point UX is way worse in (in sidebar, no need to see actions for a specific plugin of one env in all other env, nor search actions for specific document ID / collection mapping), would need to implement a better error handling with user feedback and maybe change the sidebar UI
 * First point Avoid duplicate simple actions in each env 
 * First point Allow user to export action of another env (convenient to share an action without need to change env while working on one)

I don't think that it is a problem that actions displayed when see / import-export are different 
Maybe if the import export modal display an env selector with current env by default for example you might find it less weird ?
Also, listing the pro/cons above let me think that a "global" actions folder accessible in each env could be a good feature to avoid action duplications

xbill82:
@berthieresteban I understand  your point and it makes me think that actions could be _not specific to connection_. Imagine you have three environments for the same Kuzzle app (which is the case for most of our users), then you are likely in the case of performing one same action for more than one environment, right? In this case, it is OK to go to the export menu and choose the actions from another environment and then re-import them, but... wouldn't it be more convenient to just access all the actions from the same view?

rolljee:
Closing this, as this is now quite old. This should be reopened if this feature reach the roadmap

PR #935, opened 2021-07-02
Add loader to documnent list while fetching

berthieresteban:
Adapt loader system in `Layout` to allow view of `router-view` to trigger the full container loader

Avoid display wrong messages `empty list` while fetching in `data/document/page` and trigger that parent loader

Peek 2021-07-02 15-15

### DIFF
--- a/src/components/Common/AutoFocusInput.vue
+++ b/src/components/Common/AutoFocusInput.vue
@@ -1,5 +1,6 @@
 <template>
   <b-form-input
+    autofocus
     :ref="refName"
     v-model="value"
     v-bind="$attrs"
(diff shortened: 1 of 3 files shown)


Aschen (approved):

xbill82:
This is very much appreciated, false-empty state while loading was a horrible pain and I never get the time to fix it. Now, there's one thing with this particular part of the app: the empty state is quite badly implemented (my fault), since it hides the filters, which is actually nonsense, because the UI shouldn't change that much, just because there's no data.
May I ask you to go a little bit further in this fix and move the empty state down to the list, leaving the filters always visible? This would also lead to a better implementation of the busy state, avoiding the `isEmpty && !isLoading`, because this way you'll be able to wrap the empty state _into_ the busy state, like

```
v-if loading? overlay bla bla bla
   v-if empty? empty state bla bla bla
```

(don't know if I'm clear...)

berthieresteban:
@xbill82 I enhanced the fix by moving the filters and cleaning the template but I'm not sure I fit your request, could you take a look ?

Njuelle (approved):

PR #938, opened 2021-07-02
Add realtime to document list

berthieresteban:
## List view
https://user-images.githubusercontent.com/5023426/148749617-cf31c726-8c90-4b92-b286-bafa01081ddb.mp4

## Column view
https://user-images.githubusercontent.com/5023426/148750081-e2275138-20eb-4a35-9029-2a756f998f53.mp4

## Auto-sync new documents with sorting
https://user-images.githubusercontent.com/5023426/148751864-df566dd7-a87b-4e69-8f5e-caf79ef739b1.mp4


src/components/Data/Documents/Page.vue
```diff
@@ -361,6 +366,16 @@ export default {
     }
   },
   watch: {
+    enableRealtime: {
+      async handler(value) {
+        if (value) {
+          this.fetchDocuments()
```
Aschen, line 372: Missing await

src/components/Data/Documents/Page.vue
```diff
@@ -390,6 +405,41 @@ export default {
     }
   },
   methods: {
+    async subscribeToCurrentDocs() {
+      try {
+        if (this.subscribeRoomId) {
```
xbill82, line 410: NIT: You can refactor this to a dedicated function and avoid repeating this small chunk in the watcher and in this method

Aschen (changes requested):
If an user navigate to another page then the realtime subscription is still open, it shouldn't to avoid memory leak and unwanted subscription to Kuzzle

src/components/Data/Documents/Page.vue
```diff
@@ -390,6 +405,41 @@ export default {
...
+          await this.$kuzzle.realtime.unsubscribe(this.subscribeRoomId)
+          this.subscribeRoomId = null
+        }
+        const roomId = await this.$kuzzle.realtime.subscribe(
+          this.indexName,
+          this.collectionName,
+          {
+            or: this.documents.map(d => ({ equals: { _id: d._id } }))
```
xbill82, line 418: Oh... This means we don't get to see newly created documents?

src/components/Data/Documents/Page.vue
```diff
@@ -390,6 +405,41 @@ export default {
...
+          this.realtimeNotifCallback
+        )
+        this.subscribeRoomId = roomId
+      } catch (error) {
+        this.$log.error(error)
+      }
+    },
+    realtimeNotifCallback(notif) {
+      if (notif.scope === 'out') {
+        return
```
xbill82, line 429: Should delete the document from the list, shouldn't it? Or, at least, mark the document as "DELETED" in the list (so that the user doesn't see stuff disappear under their eyes...)

src/components/Data/Documents/Page.vue
```diff
@@ -390,6 +405,41 @@ export default {
...
+      }
+      if (!['replace', 'update'].includes(notif.action)) {
+        return
+      }
+      const documents = JSON.parse(JSON.stringify(this.documents))
```
xbill82, line 434: why is this?

src/components/Data/Documents/Page.vue
```diff
@@ -390,6 +405,41 @@ export default {
...
+      const idx = documents.findIndex(d => d._id === notif.result._id)
+      documents[idx] = {
+        ...documents[idx],
+        ...notif.result._source
```
xbill82, line 438: Why don't you just mutate the concerned document? You might need a `this.$set`

xbill82 (commented):
To me, this is a very good starting point to achieve a nice feature. If we merge this PR, we should go farther by adding a sort of toast that would show up whenever the list of documents change (just like on social media, when new posts are available). The toast would contain a button that would update the list (so that changes to the list happen when the user is conscious of it).
New document would appear and deleted documents would disappear with nice animations, allowing smooth transitions in the list.

What do you think?

src/components/Data/Documents/Page.vue
berthieresteban: The goal of that feature is only to see the current listed documents being updated
I didn't want to update the list while an use is browsing it

src/components/Data/Documents/Page.vue
berthieresteban: Same as above

src/components/Data/Documents/Page.vue
berthieresteban: there is actually a $set but on the whole documents array because since it's `deep` the event wasn't spread to children props
We can discuss about that on discord next we if you want

src/components/Data/Documents/Page.vue
berthieresteban: I'll update it using lodash 
For the explanation see comment below

berthieresteban:
@xbill82  Your idea sounds good to me, we could also add an auto button to allow changes being applied without any user actions

berthieresteban:
Finally I'm not sure about your idea
I agree that we should enhance that feature by adding the documents deletion/creation and also to allow the user validate the changes 
But I'm not sure about the toast, in an asset collections for example it could happen that a lot of documents are updated at the same time so the user would be spammed of toasts...  
Maybe we should think about another solution with a better UX ?
Maybe we also could let this for now and wait for feedback that could help us improve it ?

xbill82:
@berthieresteban you're right, toast isn't the right word to describe it. You can take a look at what we did on Kuzzle IoT Platform (Alert History view). It's not actyally a "toast" but just a button that appears whenever something new happens in the collection, and it appears only once.

berthieresteban:
We discussed about the UI/UX of that feature with @Njuelle, here is our propal:

  - A `Realtime` button with a checkbox inside next to the `Refresh` button and checked by default
  - On realtime notification, an info toast spawn on bottom-right without autohide indicating that there are realtime modifications with a button to allow user to validate the modifications, clicking on it will trigger the fetch function (to handle the es sort on new updated/replaced/created/deleted documents.
  - Only one toast will be displayed at a time, no matters how many realtime notifications
  - The user will be able to dismiss the toast by clicking on the top right cross
  - Click on the realtime button to enable/disable realtime (subscription) and toast

berthieresteban:
Last proposal
ACRealtimePropal

src/components/Data/Documents/Views/Column.vue
```diff
@@ -313,6 +320,34 @@ export default {
     }
   },
   methods: {
+    getItemBadge(item) {
```
xbill82, line 323: this looks like a duplicate

Aschen:
What about this? Would someone take back the subject? I really need this, it has been quite a few times already

src/components/Data/Documents/Page.vue
```diff
@@ -398,12 +472,111 @@ export default {
...
+          ? this.formatMeta(notification.result._source._kuzzle_info)
+          : undefined
+      }
+      if (['update', 'replace'].includes(notification.action)) {
+        const documentIdx = documents.findIndex(
+          doc => doc._id === notification.result._id
+        )
+
+        if (documentIdx !== -1) {
+          documents[documentIdx] = payload
```
xbill82, line 564: I had rather updated it directly over `this.documents`

src/components/Data/Documents/Page.vue
```diff
@@ -398,12 +472,111 @@ export default {
...
+        }
+      } else if (notification.action === 'create') {
+        documents.push(payload)
```
xbill82, line 567: We should not push the new documents in the current document page, because this is a bit like trying to re-implement the logic of pagination on the frontend. This logic belongs to Elasticsearch and it's its own responsibility.

src/components/Data/Documents/Page.vue
```diff
@@ -645,6 +819,9 @@ export default {
         )
         this.documents = res.documents
         this.totalDocuments = res.total
+        if (this.enableRealtime) {
```
xbill82, line 822: Subscription should be done by default.

src/components/Data/Documents/Page.vue
```diff
@@ -823,6 +998,10 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+.cursor {
```
xbill82, line 1001: There's probably a Bootstrap class for this.

src/components/Data/Documents/Page.vue
```diff
@@ -398,12 +472,111 @@ export default {
...
+      } else if (notification.action === 'create') {
+        documents.push(payload)
+      }
+      this.$set(this, 'documents', documents)
+    },
+    realtimeNotifCallback(notif) {
+      if (!this.handledNotificationActions.includes(notif.action)) {
+        return
+      }
+      this.notifications.push(notif)
```
xbill82, line 575: I think it would be better if `this.notifications` was a hashmap where notifications are indexed by ID, like the following
```
{
  document-id: {
     ...notification
  }
}
```

xbill82 (changes requested):
@matissepage will take over this PR

Aschen (changes requested):
### New properties added after `_kuzzle_info`

If a document is updated with a new property, it appears after the `_kuzzle_info` property
<details><summary>screen</summary>
image

```
kourou collection:create test test2 '{}'
kourou document:create test test '{     
  city: "Istanbul2",
  event: "IKSV Tiyatro2"                  
}' --id document3
kourou document:update test test '{     
  city: "Istanbul2",
  event: "IKSV Tiyatro2", "toto": "adrien"
}' --id document3
```


</details>

### Behavior is shared between all windows

It's not possible to open on windows with automatic updates and another one on the same collection without (but I'm not sure it's very important

### Leftover comments in the code

I see a lot of commented code, is there a reason to keep this code?

src/components/Data/Documents/Page.vue
```diff
@@ -26,16 +28,49 @@
...
           >
+          <b-dropdown
+            class="ml-1"
+            data-cy="Refresh-dropdown"
+            split
+            variant="outline-primary"
+            :title="
+              autoSync
+                ? 'Documents are updated in real-time'
+                : 'Refresh the list to apply pending notifications'
```
Aschen, line 39: `notifications` could be confusing for users (they are used to Instagram notifications)
```suggestion
                : 'Refresh the list to apply pending changes'
```

Aschen (changes requested):
### Number of document has changed icon

An icon with a document and a "+" sign could be more explicit
image

### Toggle auto sync

When clicking on the checkbox, the pending notification should be applied and labels removed

xbill82:
Thanks for the feedback @Aschen !

## The new documents icon
I am very unsatisfied with the current solution. During these days, I've been thinking and came to the following conclusion:
* when Auto-sync is OFF, it's no such big deal to be notified about new documents, so I'll shrink the badge to a simple bullet next to the document count and keep the hovering title (and the click effect).
* when Auto-sync is ON, it is deceiving to not see new documents popping into the search results: I've spent seconds staring at the document list waiting for my new document to appear and then going like "Oh, I have to refresh".

I think it's not impossible to provide a decent mechanism that would let new documents appear on the page. Gonna work on it.

xbill82:
:point_right: @Aschen 

## New properties after `_kuzzle_info`
I really have no idea of the reasons of this behavior. The risk is that the task becomes heavy for a low ROI. I actually wish to implement a home-made JSON renderer that would improve what we currently have. Since the current one is quite crappy. Maybe, it's better to fork it and implement a new one using a json-stable-stringify, which would allow sorting the fields before rendering them.

## Shared behavior across tabs
I opened a new tab with the same URL as my first one, then unchecked the auto-sync button and I had two tabs with different behaviors.

src/components/Data/Documents/Page.vue
```diff
@@ -26,16 +28,49 @@
...
+              data-cy="Autosync-toggle"
+              :disabled="!displayRealtimeButton.includes(listViewType)"
+              @click="autoSync = !autoSync"
+            >
+              <i
+                :class="
+                  `far ${autoSync ? 'fa-check-square' : 'fa-square'} left`
+                "
+              />
+              Toggle auto-sync
```
Aschen, line 61: A check box is already a way to toggle a feature
```suggestion
              Auto-sync
```

Aschen (changes requested):
### When a document is deleted 

When a document is deleted, the pills should be clickable so the user can update the list and having the deleted document removed

<details>

image

</details>

### When a document is updated

Maybe the updated badge could be cliquable to do only a document:get action to update the document content.  
Or the same principle on the pills

xbill82:
@berthieresteban What do you think of the current result?

rolljee (approved):

Aschen (approved):

Shiranuit (approved):

PR #948, opened 2021-07-09
Persist left Sidebar size in the local storage

### LINKED ISSUE #929: Persist left Sidebar size in the local storage
Following the feedback of a customer who uses the Admin Console very heavily, it would be interesting to save in the local storage the size of the left Sidebar in order to avoid having to resize it every time we connect.

jsatge:
## What does this PR do ?

fixes #929 

Store sidebar size in the local storage.

### DIFF
--- a/src/components/Data/Layout.vue
+++ b/src/components/Data/Layout.vue
@@ -1,6 +1,14 @@
 <template>
-  <Multipane class="DataLayout Custom-resizer" layout="vertical" @paneResizeStop="registerNewPaneSize($event)">
-    <div class="DataLayout-sidebarWrapper" :style="{ width: this.paneSize }" data-cy="DataLayout-sidebarWrapper">
+  <Multipane
+    class="DataLayout Custom-resizer"
+    layout="vertical"
+    @paneResizeStop="saveNewPaneSize($event)"
+  >
+    <div
+      class="DataLayout-sidebarWrapper"
+      :style="{ width: this.paneSize }"
+      data-cy="DataLayout-sidebarWrapper"
+    >
       <treeview
         :indexName="$route.params.indexName"
         :collectionName="$route.params.collectionName"
@@ -25,7 +33,7 @@ import { Multipane, MultipaneResizer } from 'vue-multipane'
 import Treeview from '@/components/Data/Leftnav/Treeview'
 import DataNotFound from './Data404'
 import { mapGetters } from 'vuex'
-import { setLocalStorageItem, getLocalStorageItem } from './itemsStorage'
+import { setPersistedItem, getPersistedItem } from './itemsStorage'
 
 export default {
   name: 'DataLayout',
@@ -39,15 +47,13 @@ export default {
   data() {
     return {
       isFetching: true,
-      dataNotFound: false
+      dataNotFound: false,
+      paneSize: ''
     }
   },
   computed: {
     ...mapGetters('index', ['loadingIndexes', 'loadingCollections']),
     ...mapGetters('auth', ['isAuthenticated']),
-    paneSize() {
-      return `${getLocalStorageItem('paneSize')}px`;
-    },
     indexName() {
       return this.$route.params.indexName
     },
@@ -71,9 +77,9 @@ export default {
     }
   },
   methods: {
-    registerNewPaneSize(size) {
-      setLocalStorageItem('paneSize', size.scrollWidth);
-      console.log(getLocalStorageItem('paneSize'));
+    saveNewPaneSize(size) {
+      setPersistedItem('paneSize', size.scrollWidth)
+      this.paneSize = `${getPersistedItem('paneSize')}px`
     },
     async fetchIndexList() {
       try {
@@ -193,6 +199,7 @@ export default {
   },
   async mounted() {
     await this.lazyLoadingSequence()
+    this.paneSize = `${getPersistedItem('paneSize')}px`
   },
   watch: {
     $route: {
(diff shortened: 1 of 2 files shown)


src/components/Data/Layout.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <Multipane class="DataLayout Custom-resizer" layout="vertical">
-    <div class="DataLayout-sidebarWrapper" data-cy="DataLayout-sidebarWrapper">
+  <Multipane class="DataLayout Custom-resizer" layout="vertical" @paneResizeStop="registerNewPaneSize($event)">
+    <div class="DataLayout-sidebarWrapper" :style="{ width: this.paneSize}" data-cy="DataLayout-sidebarWrapper">
```
xbill82, line 3: Code formatting (if you don't do it here, it will end up in someone else's PR).
```suggestion
    <div class="DataLayout-sidebarWrapper" :style="{ width: this.paneSize }" data-cy="DataLayout-sidebarWrapper">
```

src/components/Data/Layout.vue
```diff
@@ -67,6 +71,10 @@ export default {
     }
   },
   methods: {
+    registerNewPaneSize(size) {
```
xbill82, line 74: The verb `register` in function name, conventionally means associating a function (coming from outside the scope of the program) to a hook, or something similar. I think the verb you want to use here is "save".

src/components/Data/itemsStorage.ts
```diff
@@ -0,0 +1,7 @@
+export function getLocalStorageItem(item: string) {
```
xbill82, line 1: Do we really need a function that encapsulates one line of code, consisting of calling another function with the same name, minus a point and with the same amount of parameters?
I don't think so... Here, you are just adding one layer of `import` statements without introducing any abstraction nor currying.
It is not bad to directly calling the Local Storage API from within the components 😄

xbill82 (changes requested):
The feature is OK, just a list of "philosophical" nitpicking 😄

src/components/Data/Layout.vue
```diff
@@ -44,6 +45,9 @@ export default {
   computed: {
     ...mapGetters('index', ['loadingIndexes', 'loadingCollections']),
     ...mapGetters('auth', ['isAuthenticated']),
+    paneSize() {
+      return `${getLocalStorageItem('paneSize')}px`;
+    },
```
berthieresteban, line 50: Since localStorage.get won't be reactive, I'm wondering if it is necessary to use a computed there ?

src/components/Data/Layout.vue
```diff
@@ -67,6 +71,10 @@ export default {
     }
   },
   methods: {
+    registerNewPaneSize(size) {
+      setLocalStorageItem('paneSize', size.scrollWidth);
+      console.log(getLocalStorageItem('paneSize'));
```
berthieresteban, line 76: debug leftover

berthieresteban (approved):
:+1:

src/components/Data/Layout.vue
xbill82: 👍 👍 👍 👍 👍 👍 Good catch!
Definitely right. `panelSize` should be part of the component state and get synchronized with Local Storage via component logic!

src/components/Data/itemsStorage.ts
jsatge: We talked about this with you and Nico and the conclusion was that currently we are using the local storage but maybe one day it will be a database (or anything else) and that encapsulation will be useful. So in order to make it easier in the future we decided to do it like that. If you think it's still unnecessary I can change it!

src/components/Data/Layout.vue
jsatge: I'm afraid that I don't understand clearly what you mean...can you explain to me again please ?

berthieresteban: A `computed` is usefull because the inner used properties are observed so the computed will be recalculated following it properties observers
Here `localStorage.getItem` is a call to local storage and cannot be observed so even if we use `localStorage.setItem`  the `computed` won't be updated
This is why IMO it is not so usefull to use a `computed` here
You could use a `data` to store localStorage value at `mounted` and then update it at every `localStorage.setItem`

src/components/Data/itemsStorage.ts
xbill82: Oook, now I remember, thanks for recalling. It is, indeed, a little bit overkill, but let's encapsulate all the calls to Local Storage, it's not a bad idea. But, if you want to encapsulate, you should avoid using the word "LocalStorage" in the name of the wrapper function 😅 Think about the day we switch to SQLLite: we'd have to rename all the calls to `getLocalStorageItem` to `getSQLLiteItem`, which is still a bad practice. When encapsulating, use an abstract name, like `getPersistedItem` and leave to the implementation the meaning of `Persisted`. Sounds clear?

src/components/Data/Layout.vue
xbill82: A little bit more of explanations:
* a value being _observed_ means it's part of a special VueJS object, encapsulating all the attributes in getters and setters implementing the Observable Pattern, enabling data binding into the application;
* unfortunately, this is pure VueJS internal "magic" and doesn't affect the standard API exposed by the browser (like the `localStorage` object);
* a `computed` value is a value in the scope of a VueJS component of which you define the behavior of the getter (and that has no setter): the value of the getter will be updated every time the source of this value changes (i.e. all the observable values used to compute the value), but since VueJS cannot observe `localStorage.get`, the value of the getter will not be updated if the value in the Local Storage changes.

Sounds clear? This is an advanced key concept and if you need more help, we can talk directly on Discord.

jsatge: Very clear explanations, thank you !

xbill82 (approved):

PR #951, opened 2021-07-13
 Allow to search on _id in AdvancedFilter

### LINKED ISSUE #947: Being able to search documents by ID
## Feature Description

Search documents by `_id`.

## Example Use Case

Search devices by their UUID on the SFA SAV.

## Possible Solution (optional)

Add a '_id' option in the 'Attribute' selector of the Advanced filters

berthieresteban:
Fix #947 

Allow to search on _id in AdvancedFilter
:warning: doesn't work for a partial match since _id field can only be search following `term`, `terms`, `match` and `query_string`

### DIFF
--- a/src/services/kuzzleWrapper-v1.ts
+++ b/src/services/kuzzleWrapper-v1.ts
@@ -261,91 +261,73 @@ export class KuzzleWrapperV1 {
           return
         }
 
-        // TODO Convert this into a big switch/case
-        if (filter.operator === 'contains') {
-          if (mappingAttrs[filter.attribute].type === 'text') {
-            formattedFilter.bool.must.push({
-              match_phrase_prefix: { [filter.attribute]: filter.value }
-            })
-          }
-          if (mappingAttrs[filter.attribute].type === 'keyword') {
-            formattedFilter.bool.must.push({
-              regexp: {
-                [filter.attribute]:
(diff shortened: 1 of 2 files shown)


src/services/kuzzleWrapper-v1.ts
```diff
@@ -261,91 +261,73 @@ export class KuzzleWrapperV1 {
...
-              [filter.attribute]: {
-                gt: filter.gt_value,
-                lt: filter.lt_value
+            break
+          case 'equal':
+          case 'not_equal':
+            filterToAdd = {
+              range: {
+                [filter.attribute]: {
+                  gte: filter.value,
```
xbill82, line 300: What if the type of the field is not numeric?

xbill82 (approved):

src/services/kuzzleWrapper-v1.ts
berthieresteban: I didn't change the logic in that part, I just refactored it using a switch case
you can see [here](https://github.com/kuzzleio/kuzzle-admin-console/blob/cdfd2fed39032afe167412c14110ef89e2a8ce7e/src/services/kuzzleWrapper-v1.ts#L293) that the 3 `else if` of `equal` `not_equal` and `range` where using a ES `range` filter so I let that filter.
If we need to change that, I would prefer doing it in another PR

Njuelle (approved):

PR #953, opened 2021-07-13
Add feedback dropdown

### LINKED ISSUE #913: Add a feedback form
## Feature Description
We need to collect feedback about the features implemented in the Admin-console.

### Possible solution
Rating with 5 stars on  admin console pages

Also a form with simple text?

berthieresteban:
fix #913 

Add a main menu entry with a dropdown to allow users give us feedback on discord/stackoverflow/github

image

### DIFF
--- a/src/components/Common/MainMenu.vue
+++ b/src/components/Common/MainMenu.vue
@@ -45,16 +45,26 @@
         >
           API Action
         </b-nav-item>
-        <b-nav-text
-          class="ml-2"
-          title="Give us your feedback on our GitHub page"
+        <b-dropdown
+          class="mr-2"
+          no-caret
+          :toggle-class="
+            `feedbackDropdown EnvColor--${currentEnvironmentColor}`
+          "
         >
-          <a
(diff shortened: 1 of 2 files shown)


src/components/Common/MainMenu.vue
```diff
@@ -89,6 +99,29 @@ export default {
   components: {
     EnvironmentSwitch
   },
+  data() {
+    return {
+      feedbackOptions: [
```
xbill82, line 104: Why factorizing the content of the dropdown in the component state? Is it meant to change dynamically based on user interaction?

xbill82 (approved):
👍

src/components/Common/MainMenu.vue
berthieresteban: IMHO it allow a better readability also I think some data used for component rendering are totally legitimate to be there even if they are not dynamic
But I can just put it in the template if needed

Njuelle (approved):

PR #959, opened 2021-07-28
Add dropdown-view to realtime document view

### LINKED ISSUE #957: Missing dropdown view type in realtime view
When browsing documents, if we change the view to realtime view we can't change view type anymore

acviewtype

berthieresteban:
fix #957 

fix dropdown-view selector in realtime document view

## Boyscout

Fix map shapes

### DIFF
--- a/src/components/Data/Documents/Views/Map.vue
+++ b/src/components/Data/Documents/Views/Map.vue
@@ -334,17 +334,17 @@ export default {
     },
     circleShapes() {
       return this.shapesDocuments.filter(
-        shape => shape.content.type === 'circle'
+        shape => shape.content && shape.content.type === 'circle'
       )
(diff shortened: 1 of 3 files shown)


Njuelle (approved):

src/components/Data/Documents/Views/Map.vue
```diff
@@ -334,17 +334,17 @@ export default {
     },
     circleShapes() {
       return this.shapesDocuments.filter(
-        shape => shape.content.type === 'circle'
+        shape => shape.content && shape.content.type === 'circle'
```
xbill82, line 337: Better use Lodash's `get` here

xbill82 (changes requested):

PR #968, opened 2021-08-16
redirect to Login page if token expired

### LINKED ISSUE #955: Unable to logout if token is invalid
If the token is invalid (which happens if you restart your backend) the logout action fails and the login page is never displayed until the page is refreshed.

alexandrebouthinon:
Fix #955 

## What does this PR do ?

As explained in the issue, as it we're unable to logout (to choose another environment) if token expired or if `auth.logout` action errored. 

- Run a Kuzzle stack
- Run this PR Admin Console version
- Connect to your Kuzzle and create a first admin
- Find your Kuzzle docker container and stop it (`docker stop myContainer`)
- Restart it
- Now, it should redirect you to the current environment login page (and allow you to choose another environment)


xbill82 (changes requested):
This implementation is OK, but I'd like to take the toast out: it is useless and confusing for the user to be visually notified that something went wrong, once they're on the login page.

xbill82 (approved):

Njuelle (approved):

src/components/Common/MainMenu.vue
```diff
@@ -147,20 +147,10 @@ export default {
     async doLogout() {
       try {
         await this.$store.direct.dispatch.auth.doLogout()
-        this.$router.push({ name: 'Login' })
```
xbill82, line 150: Why this is deleted?

alexandrebouthinon: I use a `finally` statement because we want to redirect users to the login page either logout succeed or failed

xbill82: oh, right!

xbill82 (approved):

PR #976, opened 2021-10-29
Bump Kuzzle SDK version

xbill82:
Uses latest version of JS SDK.


Aschen:
@xbill82 Since you updated the SDK, you should use the [authenticator](https://docs.kuzzle.io/sdk/js/7/core-classes/kuzzle/properties/#authenticator) property to handle authentication so the SDK will automatically refresh token and also re-auth if the connection is lost

PR #979, opened 2021-11-04
[feature] Export data to CSV

xbill82:
You waited long for it, it is here. This PR allows to export items currently shown in a table view as a CSV file.
Ok, this is a first iteration, it might need more varnishing to make everyone happy.

Kapture 2021-11-05 at 11 58 26

## Other improvements

* Column view needed some serious love
* MOAR tests on Column view

### DIFF
--- a/src/components/Data/Documents/Views/Column.vue
+++ b/src/components/Data/Documents/Views/Column.vue
@@ -200,10 +200,10 @@
                 :id="`col-${item._id}-${field}`"
               >
                 <template v-if="item[field] === null">
-                  <b-badge>null</b-badge>
+                  <code>null</code>
                 </template>
                 <template v-else-if="item[field] === undefined">
-                  <b-badge>undefined</b-badge>
+                  <code>undefined</code>
                 </template>
                 <template v-else-if="Array.isArray(item[field])">
                   <b-badge
--- a/src/components/Data/Documents/Views/Column.vue
+++ b/src/components/Data/Documents/Views/Column.vue
@@ -7,7 +7,7 @@
           data-cy="SelectField"
           variant="outline-secondary"
           menu-class="dropdownScroll"
-          text="Select columns to display"
+          text="Fields"
           no-flip
         >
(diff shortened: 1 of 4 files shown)


jenow:
That's a great addition but I don't think it suits most of users needs.
If we can't export the whole collection that's very limitating. I know the feature has just been added to this existing view with pagination but it would be better to maybe have a specific view for this which can actually be usable with date pickers to select a time range. Adding features is great but they have to actually be useful.
In the meantime I think adding a checkbox "export whole collection" and date pickers would make this feature make more sense from a user point of view

xbill82:
@jenow yes, your statement makes sense. We are currently thinking about the next iterations on this feature and will take your ideas into account.

rolljee (approved):

src/components/Data/Documents/Views/Column.vue
```diff
@@ -313,6 +324,39 @@ export default {
     }
   },
   methods: {
+    isObject: _.isObject,
+    promptExportCSV() {
+      this.$bvModal
```
Njuelle, line 329: Nitpicking: you can `await` the confirmation modal instead of attaching a `then` callback

Njuelle (approved):
I tried a massive export, the file is OK (however it is all messed up in "LibreOffice" idk why).

But I think we'll have to give the column view a boost before we can release this feature.
When I select a lot of fields to display in the table (from 20 fields), the admin console becomes very slow or even unusable

xbill82:
@Njuelle I actually took a look at the problem and it seems very costly to solve this problem. With 500 rows, the page becomes very bloated, as you state (regardless of the number of selected fields). My feeling is that the bottleneck is the bootstrap table component. Do you think the ROI of blocking this feature while we dig into that component (or something else) is interesting? Shouldn't we rather release this and let the feedback come to us?

Njuelle:
@xbill82 Indeed, let's make a release of this feature (at least on next-console) and deal with the perf problem on large dataset in a second time

PR #981, opened 2021-12-01
Populate collection edition form with complete mappings

Aschen:
## What does this PR do ?

The collection edition form was populating it's content with directly the mappings `properties` despite the API expect to receive a full mappings object (with top level properties: `dynamic`, `properties`)

### Other changes

  - display formatted dates on other view

### DIFF
--- a/src/components/Data/Collections/Update.vue
+++ b/src/components/Data/Collections/Update.vue
@@ -7,7 +7,7 @@
         submit-label="Update"
         :collection="collection.name"
         :index="index.name"
-        :mapping="getFullMappings"
+        :mapping="fullMappings"
         :realtime-only="collection.isRealtime()"
         @submit="update"
       />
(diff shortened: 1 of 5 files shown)


src/components/Data/Collections/Update.vue
```diff
@@ -53,10 +53,13 @@ export default {
         this.collectionName
       )
     },
-    mappingAttributes() {
-      return this.collection
-        ? omit(this.collection.mapping, '_kuzzle_info')
-        : null
+    getFullMappings() {
```
Njuelle, line 56:
```suggestion
    fullMappings() {
```
Nitpicking: a computed function is always a getter function, you don't need to specify a verb in its name

Njuelle (approved):

xbill82 (approved):

PR #987, opened 2022-02-21
Use Kepler to retrieve product usage metrics

alexandrebouthinon:
## What does this PR do ?

As we done on Kourou, let's use it on the Admin Console

## ToDo 
- [x] Integrate Kepler Companion in the Admin Console
- [x] Make a "cookie banner" to permit users to disable it (@Njuelle I'll need your help on this) 
- [x] Add current environment name `console.kuzzle.io` or `next-console.kuzzle.io` in Kepler tags


src/main.ts
```diff
@@ -21,8 +22,17 @@ Reflect.defineProperty(window, 'kuzzle', {
 Vue.use(BootstrapVue)
 Vue.use(BootstrapVueIcons)
 
+const analytics = new KeplerCompanion()
 const router = createRoutes(Vue.prototype.$log)
 
+router.afterEach(async (to, _) => {
+  await analytics.track({
```
Aschen, line 29: You should not await here but only empty catch to not degrade the UX

Aschen (approved):

src/routes/index.ts
```diff
@@ -152,5 +154,24 @@ export default function createRoutes(log) {
...
 
+  const analytics = new KeplerCompanion()
+  router.afterEach(async (to, _) => {
+    const shouldAddTelemetry =
+      telemetryCookies.get() === null || telemetryCookies.get() === 'false'
+        ? false
+        : true
+
+    if (shouldAddTelemetry) {
+      await analytics.add({
```
Aschen, line 165: Can you add a catch to this promise so it never prints error to the console?

Aschen (changes requested):

Njuelle (approved):

Njuelle:
the tests fail because the new banner covers elements of the DOM and prevents the click on these elements. IMHO the best way to fix this would be to disable the banner and the telemetry during the tests

Aschen (approved):

Aschen:
LGTM :+1:

PR #993, opened 2022-11-08
Refactor of the export csv button to export more than 500 documents max 

thomas-mauran:
##  Refactor of the export csv button to export more than 500 documents max 

Refactoring the export csv function to know export more than 500 documents max 

image

  - Step 1 : open a kuzzle backend with data
  - Step 2 : go on a collection with a lot of document
  - Step 3 : try the export button

### Other changes

deleted the function  convertToCSV() from src/services/cllectionHelper.ts


src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -211,7 +210,7 @@
         </b-table-simple>
       </b-col>
     </b-row>
-    <export-csv-modal
+    <!-- <export-csv-modal
```
Aschen, line 213: Comments should be removed

src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -240,13 +239,13 @@ import HighlightableRow from './HighlightableRow.vue'
 import PerPageSelector from '@/components/Common/PerPageSelector'
 import NewDocumentsBadge from '../../Common/NewDocumentsBadge.vue'
 
-import { convertToCSV, flattenObjectMapping } from '@/services/collectionHelper'
-import ExportCsvModal from '@/components/Data/Documents/Common/ExportCSVModal'
+import { flattenObjectMapping } from '@/services/collectionHelper'
+// import ExportCsvModal from '@/components/Data/Documents/Common/ExportCSVModal'
```
Aschen, line 243: Same remark about comments

src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -353,23 +353,43 @@ export default {
...
-      a.download = `${this.index}-${this.collection}.csv`
-      a.href = URL.createObjectURL(blob)
-      a.addEventListener('click', () => {
-        setTimeout(() => URL.revokeObjectURL(a.href), 30 * 1000)
+
+      // Request to the export api endopoint
+      let request = new XMLHttpRequest()
+      request.open(
+        'POST',
+        `http://localhost:7512/${this.index}/${this.collection}/_export?format=csv`
```
Aschen, line 368: You need to retrieve the Kuzzle host and port from the SDK instance

src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -353,23 +353,43 @@ export default {
...
+      )
+      request.setRequestHeader('Content-Type', 'application/json')
+      request.responseType = 'text'
+
+      // Request body with the selected fields and the current querry
```
Aschen, line 373:
```suggestion
      // Request body with the selected fields and the current query
```

src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -353,23 +353,43 @@ export default {
...
+      let body = JSON.stringify({
+        query: this.searchQuery,
+        fields: this.selectedFields,
       })
-      a.click()
+
+      // Send the request
+      request.send(body)
+
+      let filename = `${this.index}-${this.collection}-export.csv`
```
Aschen, line 382:
```suggestion
      const filename = `${this.index}-${this.collection}-export.csv`
```

src/components/Data/Documents/Views/Column/Column.vue
```diff
@@ -353,23 +353,43 @@ export default {
...
+      let request = new XMLHttpRequest()
+      request.open(
+        'POST',
+        `http://localhost:7512/${this.index}/${this.collection}/_export?format=csv`
+      )
+      request.setRequestHeader('Content-Type', 'application/json')
+      request.responseType = 'text'
+
+      // Request body with the selected fields and the current querry
+      let body = JSON.stringify({
```
Aschen, line 374:
```suggestion
      const body = JSON.stringify({
```

Aschen (changes requested):
Can you improve the PR title and also assign it to you please?

Aschen (approved):

sebtiz13:
Export doesn't work with huge collection (example: more than 10k documents).
The actual solution isn't really good, because the data are stored in client memory before conversion in blob to create download link, this can cause block the JS browser runtime, also this solution has an UX problem because with a large number of documents, user doesn't have indication of status (loading status or progress).

I think an better client solution it's to use Web Worker for non blocking process an split treatment in chunk.
IMHO the best solution, it's to implement a stream solution on backend and client just open a link, so the browser treat the download progress independently.

Shiranuit:
> Export doesn't work with huge collection (example: more than 10k documents). The actual solution isn't really good, because the data are stored in client memory before conversion in blob to create download link, this can cause block the JS browser runtime, also this solution has an UX problem because with a large number of documents, user doesn't have indication of status (loading status or progress).
> 
> I think an better client solution it's to use Web Worker for non blocking process an split treatment in chunk. IMHO the best solution, it's to implement a stream solution on backend and client just open a link, so the browser treat the download progress independently.

Using the export method is the proper way to download large amount of documents directly from the backend, this method is able to handle extremely large amount of data by streaming them directly from the backend, you should be able to download the file directly by letting the browser open the following url `https://<kuzzle host>/<index>/<collection>/_export` and configure every options of the export by putting things into the query params (lang, searchBody, scroll etc...)
The only thing I'm not sure how to do is how to provide the authentication token when going to the export url

sebtiz13:
@Shiranuit It's not possible to provide the authentication token in GET params of url ?
If it's not possible due to security reason (understandable), one solution can be add an method to generate one temporary token (this token can contain the params of export) to provide in request. This temporary token are valid just one time.
IMHO it's the better solution to export securely without allow export spamming.

Aschen:
> The only thing I'm not sure how to do is how to provide the authentication token when going to the export url

If the SDK use the cookie auth then the auth cookie should be forwarded automatically no?

Shiranuit:
> @Shiranuit It's not possible to provide the authentication token in GET params of url ? If it's not possible due to security reason (understandable), one solution can be add an method to generate one temporary token (this token can contain the params of export) to provide in request. This temporary token are valid just one time. IMHO it's the better solution to export securely without allow export spamming.

@sebtiz13 Actually it's not possible to provide the authentication token directly in the URL, also it's not recommended even though query params are encrypted, the proper solution would be with cookies

> > The only thing I'm not sure how to do is how to provide the authentication token when going to the export url
> 
> If the SDK use the cookie auth then the auth cookie should be forwarded automatically no?

@Aschen Yes it might work with the cookie but this will require the admin console to support login with cookies which is not the case right now, I dont think this will be that hard to implement

Aschen:
@Shiranuit I think cookie auth should be default, isn't it the case ?

Otherwise I agree but if the backend didn't enable cookies I don't see how we can properly authenticate the client..

sebtiz13:
Ok like I say it's understandable not to be able to provide directly token in GET parameters, but what you think of solution of generate a temporary token for export.
This token would only be valid for one export and not use the user token. So, off course this create a specific case, but don't cause security problem and it's simpler than cookie.

Aschen:
@sebtiz13 We could do this but it require some backend development for Kuzzle

src/components/Data/Documents/Views/Column/Column.vue
alexandrebouthinon: Maybe we should put this HTTP call in the store (since it's a backend interaction) to access the SDK instance and reuse values from it:

This is only an example not sure the name of the variables:
```suggestion
      // Request to the export api endopoint
      let request = new XMLHttpRequest()
      request.open(
        'POST',
        `${this.sdk.protocol.ssl ? 'https' : 'http'}://{this.sdk.protocol.host}:${this.sdk,protocol.port}/${this.index}/${this.collection}/_export?format=csv`
      )
      request.setRequestHeader('Content-Type', 'application/json')
      request.setRequestHeader('Authorization', `Bearer ${this.sdk.jwt}`)
      request.responseType = 'text'

      // Request body with the selected fields and the current querry
      let body = JSON.stringify({
```

What do you think @Aschen @Shiranuit @sebtiz13?

sebtiz13: I think it's not really useful to move this on the store, because this action doesn't need to save the result.
IMHO, the store should be seen like a local storage for browser, to store data we need to share between different pages and components.

alexandrebouthinon: I agree with you about the move the Vue.js store. But what do you think about reuse all the SDK info (including the current JWT)?

alexandrebouthinon: I think it brings a solution on the discussion of @Shiranuit  and @Aschen  about the need to pass the token in the URL

Aschen: This PR is stil WIP. The code you are talking about has been removed anyway

thomas-mauran:
image

Here is what the css looks like

